PR #1, opened 2021-12-13
Several minor bugfixes for local allocations

stedolan:
- Signature inclusion between `external` definitions was incorrect, which broke when an `external` definition in an `ml` file is matched against a less general `external` declaration in an `mli`
  - Selectgen failed to close regions after a try in tail position. (It now uses the same code as everything else, removing the special case for try)
  - Local domain state is now initialised properly
  - Functions with no cases (for which there is no syntax) are now supported again, as certain ppxs generate them

(Best reviewed commit-by-commit)


mshinwell (approved):
Approving for the Selectgen and domain state initialisation changes.

antalsz (approved):

PR #3, opened 2022-02-01
Better tail call behaviour in caml_applyN

stedolan:
`caml_applyN` needs a region, but this can prevent tail-call optimisation in some cases. This patch adds a runtime check to the slowpath, preserving tail calls when possible. (The new test in `partial.ml` results in a stack overflow without this change)


lpw25 (approved):
LGTM

PR #7, opened 2022-03-02
Local solver speedups

stedolan:
The local modes solver is slow sometimes, on e.g. `testsuite/fma/fma.ml`. Most of the time is spent in `check_const`, computing the transitive closure of mode variables.

This PR contains two independent fixes for this. Either is sufficient to fix the compile performance problem, but I think we should do both:

  1. Remove the linear-time `List.memq` checks in `compress_vlower`, in favour of a constant-time mark bit.
  2. Do not call `check_const` in `register_allocation`. `check_const` does a full transitive closure, which can be expensive and should not be done eagerly in a speculative optimisation.

The implementation of (1) uncovered some bugs in the snapshot / backtrack handling in `compress_vlower`, so it got refactored somewhat.


typing/btype.ml
```diff
@@ -1005,21 +1009,58 @@ module Alloc_mode = struct
...
         trans_low v'
       end
     and trans_low v' =
-      submode_exn (Amode v'.lower) (Amodevar v);
+      assert (v != v');
+      if not (le_const v'.lower v.upper) then
+        Misc.fatal_error "compress_vlower: invalid bounds";
+      if not (le_const v'.lower v.lower) then begin
+        let m = join_const v.lower v'.lower in
+        set_lower ~log v m;
```
lpw25, line 1044: Since we're not editing `v.vlower` directly, it might be clearer to have a `new_lower` reference too, and make all the changes to `v` right at the end.

lpw25 (approved):
LGTM

typing/btype.ml
stedolan: Good idea, fixed.

PR #8, opened 2022-03-02
Remove extra modevars

stedolan:
Removes some unnecessary mode variables (On top of #7)


lpw25 (approved):
These all look fine. The reasoning is that we only care about the mode of expressions for expressions that allocate, and in none of the cases that we are making local can there be an allocation. I've checked the code in lambda, and we do indeed ignore the mode of expressions other than potentially allocating expressions -- modulo a couple of assertions which should still be fine.

Note that "potentially allocating" isn't quite the same as "allocating": we always look at the mode of function applications since they will allocate if they are partial, even when the type tells us that they can't be partial. I think this means that this patch can cause us to generate slightly different code -- using the local versions of caml_apply in more places -- but only in places where we can't observe the difference.

stedolan:
This had nonzero effect on generated code, for the reason you mentioned: the modes of some applications changed. This changes whether the surrounding function's region can be elided.

For now, I've shrunk this patch down to contain only the change to argument modes. This one is safe (the only change is to avoid creating variables which will be immediately constrained to bot), and by itself accounts for ~60% of the benefit of this patch (measured on typecore).

lpw25 (approved):

PR #9, opened 2022-03-07
Accept [@ocaml.local] without -extension

stedolan:
The autogenerated attributes are renamed to [@extension.foo]


typing/typedecl.ml
```diff
@@ -202,12 +202,12 @@ let make_params env params =
...
 let has_global_attr attributes =
   List.exists
-    (fun attr -> String.equal attr.attr_name.txt "ocaml.global")
+    (fun attr -> String.equal attr.attr_name.txt "extension.global")
     attributes
 
 let has_nonlocal_attr attributes =
   List.exists
-    (fun attr -> String.equal attr.attr_name.txt "ocaml.nonlocal")
+    (fun attr -> String.equal attr.attr_name.txt "extension.nonlocal")
```
lpw25, line 210: Can we move these functions to `Builtin_attributes`? It would be good to only need to change one place

lpw25 (commented):
This makes `extension.local` valid everywhere that previously used only `ocaml.local`. However, the expression version still only supports `extension.local` as an extension node.

We need to be careful here to avoid `ocaml.local` overlapping with the annotation to control the "local functions" optimisation, but we do need something so that we can make local returning functions in public-release libraries. How about `[@as local]` in expressions? or maybe `[@allocate local]`?

typing/typedecl.ml
lpw25: Also this should support both `ocaml.global` and `extension.global`

PR #10, opened 2022-03-07
Improve inclusion error messages for [@local_opt]

stedolan:
Previously, `[@local_opt]` never appeared in error messages, making some of them very confusing.


lpw25 (approved):
Maybe rename `attr_of_native_repr`, but otherwise good to go

PR #13, opened 2022-03-31
Improve the API of language extensions to better support upstream compatibility (and also tooling)

antalsz:
Carl wants a better interface to programmatically work with language extensions for refactoring tools: in particular, he wants to be able to temporarily en/disable language extensions.  A weaker desire was to have a non–string-based API, and to make clear that the function formally known as `disable_all` is rather more permanent than just disabling all extensions.  I cleaned up this API, and wrote some test cases for it.


stedolan (changes requested):
lgtm, modulo some invariant nitpicking

stedolan (approved):

antalsz:
I'd like to revive this PR!  I think this is a better UI and will serve us better going forwards.  It currently has a few issues:

1. The name `Universe` is terrible (but at least currently isn't user-facing).
2. The (status quo) error message when you specify `-disable-all-extensions` or `-only-erasable-extensions` and then also `-extension FOO` is misleading; it tells you to specify `-extension FOO`, but of course sometimes you can't.  This isn't a regression, but could be improved with a slightly different design here.
3. Because `local_` now always parses and only fails during type-checking, I have a test that doesn't make sense anymore.  Not sure if the right thing to do is to move the test over to comprehensions (which fails during parsing) or modify `local_`; thoughts?
4. We need tests for the new `-only-erasable-extensions` option; I'm waiting to see what makes sense for the prior point.

In a future PR, I'll also be moving `Extensions` to its own file, which I think is going to make life easier for ocamlformat and Merlin, but I haven't done that here so as to preserve the diff.

goldfirere:
As usual for me, before implementation comes specification: what's the goal of this patch? Reading through, it looks like this patch changes two APIs:

1. It changes the programmable API, as used by e.g. an executable that links with compilerlibs. This new API allows callers to restrict the set of possible extensions, and then allows you to enable and disable extensions within that set (perhaps locally).
2. It changes the user-facing API to allow `-only-erasable-extensions`.

Some observed effects of the current design:
* Both programmatically and from the command-line, it is only ever possible to restrict the universe, never to expand it.
* It seems that `-extension comprehensions -only-erasable-extensions` will work (but not allow comprehensions), while `-only-erasable-extensions -extension comprehensions` will error. I suppose that's OK, but is this the expected behavior?
* There is no way from the command line to have, say, only include-functor enabled, but nothing else.

I suppose for our known use cases, this interface is sufficient. But actually, for our known use cases, I don't see a reason to maintain e.g. include-functor and polymorphic-parameters as different extensions. That is, maybe we should just have `-extension erasable` and `-extension all` and `-disable-extensions`?

Again, I'm not actually against the design as included here, but I want it surfaced so we can make sure we're happy with it. (And then, staying in character, I would want it documented in the manual -- at least the command-line part of the interface. The programmer API is indeed documented in the mli file.)

antalsz:
I added `-no-extension` and tried to update the manual but couldn't build it (I don't think we can on our machines since we don't have a good TeX, but maybe that wasn't the problem we had).  We agreed the behavior of universe-restriction made sense once we talked about it.  There's just one more thing I need to examine and then we should be good!

antalsz:
Good to go, I think!  @goldfirere?

driver/main_args.ml
```diff
@@ -731,15 +731,32 @@ let mk_extension f =
...
 let mk_disable_all_extensions f =
   "-disable-all-extensions", Arg.Unit f,
-  " Disable all extensions, wherever they are specified; this flag\n\
-  \    overrides the -extension flag (whether specified before or after this\n\
-  \    flag), disables any extensions that are enabled by default, and\n\
-  \    ignores any extensions requested in OCAMLPARAM."
+  " Disable all extensions, wherever they have been specified; this\n\
+  \    flag overrides prior uses of the -extension flag, disables any\n\
+  \    extensions that are enabled by default, and causes future uses of\n\
+  \    the -extension or -no-extension flag to raise."
```
goldfirere, line 750:
```suggestion
  \    the -extension or -no-extension flag to error."
```

Why `-no-extension`? I think `-no-extension` should be allowed even when extensions are disabled.

driver/main_args.ml
```diff
@@ -731,15 +731,32 @@ let mk_extension f =
...
+;;
+
+let mk_only_erasable_extensions f =
+  "-only-erasable-extensions", Arg.Unit f,
+  " Disable all extensions that cannot be \"erased\" to attributes,\n\
+  \    wherever they have been specified; this flag overrides prior\n\
+  \    contradictory uses of the -extension or -no-extension flags,\n\
```
goldfirere, line 757: Well, it doesn't override prior uses of `-no-extension`.

driver/main_args.ml
```diff
@@ -731,15 +731,32 @@ let mk_extension f =
...
-  \    ignores any extensions requested in OCAMLPARAM."
+  " Disable all extensions, wherever they have been specified; this\n\
+  \    flag overrides prior uses of the -extension flag, disables any\n\
+  \    extensions that are enabled by default, and causes future uses of\n\
+  \    the -extension or -no-extension flag to raise."
+;;
+
+let mk_only_erasable_extensions f =
+  "-only-erasable-extensions", Arg.Unit f,
+  " Disable all extensions that cannot be \"erased\" to attributes,\n\
```
goldfirere, line 755: I think this should just list the extensions in play.

manual/src/cmds/unified-options.etex
```diff
@@ -851,5 +851,27 @@ hyphen (-).}
...
+either with the same or a different language extension; is idempotent.
+
+\item[(JST) "-no-extension" \var{language-extension}]
+Disable the specified \var{language-extension}.  Can be specified more than once,
+either with the same or a different language extension; is idempotent.
+
+\item[(JST) "-only-erasable-extensions"]
+Restricts the "-extension" and "-no-extension" options to work only with
+so-called ``erasable'' extensions: ones that can be rewritten into attributes
+while still preserving the program's observable runtime behavior.  Turns off
```
goldfirere, line 865: There's a lot going into "observable" here. The amount of allocations is an observable quantity, for a sufficiently astute observer. Maybe change "observable" to "input/output"?

testsuite/tests/language-extensions/language_extensions.ml
```diff
@@ -0,0 +1,163 @@
+(* TEST
+   include ocamlcommon
+   flags = "-I ${ocamlsrcdir}/parsing"
+   reference = "${test_source_directory}/reference.txt"
+   * skip
```
goldfirere, line 5: What's going on here? This is a big new test, but it's skipped. I did not read it.

utils/clflags.ml
```diff
@@ -397,53 +399,123 @@ module Extension = struct
...
+     (1) [!extensions] contains at most one copy of each extension.
+
+     (2) Every member of [!extensions] satisfies [Universe.is_allowed !universe].
+         (For instance, [!universe = No_extensions] implies
+         [!extensions = []]). *)
+  let extensions = ref default_extensions (* -extension *)
+  let universe   = ref Universe.Any       (* -only-erasable-extensions,
+                                             -disable-all-extensions *)
+
+  let set extn ~enabled =
```
goldfirere, line 461: As I suggested elsewhere, I think `-no-extension` should always be OK.

utils/clflags.ml
```diff
@@ -397,53 +399,123 @@ module Extension = struct
...
-    not !disable_all_extensions
-    && (List.mem ext default_extensions
-        || List.mem ext !extensions)
+        "Cannot %s extension %s: incompatible with %s"
+        (if enabled then "enable" else "disable")
+        (to_string extn)
+        (Universe.compiler_options !universe)));
+    if enabled then begin
+      if not (List.exists (equal extn) !extensions) then
+        extensions := extn :: !extensions
```
goldfirere, line 470: Should this just be a set instead of a list? Maybe not, because it's so small, so constant factors overwhelm asymptotic ones. In this case, though, just use `filter` below, without the early cutoff. That is, either this is small and simple, or it's big and needs to be efficient. Not sure which is better, but I think small and optimized is a poor choice.

utils/clflags.ml
```diff
@@ -397,53 +399,123 @@ module Extension = struct
...
+
+  let is_enabled extn = List.mem extn !extensions
+
+  (* It might make sense to ban [set], [enable], [disable],
+     [only_erasable_extensions], and [disallow_extensions] inside [f], but it's
+     not clear that it's worth the hassle *)
+  let with_set extn ~enabled f =
+    let current_extensions = !extensions in
+    let current_universe   = !universe   in
+    Fun.protect
```
goldfirere, line 491: I think this function basically just inlines `Misc.protect_refs`. Maybe just use that instead?

driver/main_args.ml
antalsz: I can see this both ways.  Right now, this directly calls `disable`, which also fails, and I'm more inclined to believe _that_'s the right behavior.

driver/main_args.ml
antalsz: I changed this to

```
let mk_only_erasable_extensions f =
  let erasable_extensions =
    let open Clflags.Extension in
    all |> List.filter is_erasable |> List.map to_string |> String.concat ", "
  in
"-only-erasable-extensions", Arg.Unit f,
  " Disable all extensions that cannot be \"erased\" to attributes,\n\
  \    wherever they have been specified; this flag overrides prior\n\
  \    contradictory uses of the -extension or -no-extension flags,\n\
  \    raises on future uses of them, and disables any such extensions\n\
  \    that are enabled by default.\n\
  \    (Erasable extensions: " ^ erasable_extensions ^ ")"
;;
```

Is that what you meant?

manual/src/cmds/unified-options.etex
antalsz: e64e0bde88

utils/clflags.ml
antalsz: I think you are probably correct, so I've made the change, but it galls me slightly because it's so obviously inefficient.  (By, like, 2 things.  In a case which will ~never be called.)  Leaving this comment to get that out of my system and making the code change.

utils/clflags.ml
antalsz: Not quite?  `Misc.protect_refs` sets them directly, which we don't want to do here.

testsuite/tests/language-extensions/language_extensions.ml
antalsz: The reason is on the line below!  I completely forgot about this.  There should be a comment somewhere but I'm not sure where it went.  This is the danger in making commits marked "Temporary".  What's going on here: I wrote this test back when `local_` only parsed in the presence of `-extension local`.  Now, it always parses and fails later.  There are two changes we can make: (1) make `local_` only parse in the presence of `-extension local` again, or (2) use a modular extension language extension.  I think I'll make the change to (2), but then it's really only testing the modular extension machinery; I think that's probably fine.

antalsz: This should be reviewable now

driver/main_args.ml
antalsz: Nah, I'm convinced.

antalsz:
Thanks for the comments; give the last few changes a look-see and we can hopefully get this merged tomorrow!

antalsz:
Well, no, that's a bit of a lie.  I did the rebase and now the test fails, because with the new extensions machinery, you don't get an exception during parsing if the extension isn't enabled anymore.  Is that intended?  You still get an exception, but it shows up later (I think just when we typecheck it and try to match on it with `ast_of`).

testsuite/tests/language-extensions/language_extensions.ml
```diff
@@ -2,118 +2,148 @@
    include ocamlcommon
    flags = "-I ${ocamlsrcdir}/parsing"
    reference = "${test_source_directory}/reference.txt"
-   * skip
-   reason = "local_ currently always parses"
 *)
 
+(* Currently, this test case only works with modular extensions, as they fail
```
goldfirere, line 7: This comment seems at odd with your GitHub comment which said that modular extensions *don't* fail during parsing...

testsuite/tests/language-extensions/language_extensions.ml
```diff
@@ -2,118 +2,148 @@
...
+   during parsing and we test via parsing. *)
+
```
goldfirere, line 9: Skimming through, I don't think it's quite worth a careful review of these tests. It looks broadly that they're working and that we will discover if they're not.

goldfirere:
> e. I did the rebase and now the test fails, because with the new extensions machinery, you don't get an exception during parsing if the extension isn't enabled anymore. Is that intended? You still get an exception, but it shows up later (I think just when we typecheck it and try to match on it with `ast_of`).

Yes, that's intended. I suppose this decision could be reversed, but my general thought is that it's better to fail later in the pipeline, as parser errors are often challenging to deal with.

testsuite/tests/language-extensions/language_extensions.ml
antalsz: Yes, that's because I caught that during rebase and punted on the best way to resolve it

antalsz:
Alright, fixed the test to use typechecking instead.  Are we ready to merge once CI completes?

goldfirere (approved):

PR #16, opened 2022-04-14
Allow warning 68 to be controlled by attributes

lpw25:
Remembers whether warning 68 was enabled when we type a function so that the warning can be controlled by `[@warning]` attributes. I stored the information as part of the existing `partial` information, which is related, so keeping the data won't take any extra space in e.g. cmt files.

### DIFF
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -173,23 +173,23 @@ type binding =
   | Bind_value of value_binding list
   | Bind_module of Ident.t * string option loc * module_presence * module_expr
 
-let rec push_defaults loc bindings cases partial =
+let rec push_defaults loc bindings cases partial warnings =
   match cases with
     [{c_lhs=pat; c_guard=None;
       c_rhs={exp_desc =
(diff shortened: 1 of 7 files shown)


stedolan:
The `func_partial` type is just too weird. Wouldn't this make much more sense as two fields, one for whether the pattern is partial and one for the state of w68? Is the increase in cmt size noticeable?

lpw25:
It makes more sense, but I was hoping to avoid testing whether the space mattered or bumping the cmt magic number, since matching on this field as if it was `partial` will probably just continue to work. If we're going to waste the space anyway I will probably just store a pointer to the warnings state in there.

stedolan:
> It makes more sense, but I was hoping to avoid testing whether the space mattered or bumping the cmt magic number, since matching on this field as if it was partial will probably just continue to work

You can get the same 'probably just continue to work' behaviour by adding a new field to the _end_ of the function record.

> If we're going to waste the space anyway I will probably just store a pointer to the warnings state in there.

That sounds good

antalsz (approved):

PR #17, opened 2022-04-19
Fix a GC bug in local stack scanning

stedolan:
A one-character typo in the local GC causes the stack pointer was computed relative to the wrong local arena, leading to spurious fatal errors.

As a consistency check, the local allocations GC verifies that pointers point only to older local allocations, by comparing older and newer stack pointers. However, the calculation of the new stack pointer incorrectly used the _location_ of the field being scanned rather than its _value_, which gave the wrong answer. (This wasn't found earlier because it mostly has no effect, as having multiple live arenas with the right pattern of pointers is relatively rare)


lpw25 (approved):
LGTM

PR #18, opened 2022-04-20
Enable the local keywords even when the local extension is off

lpw25:
Avoid having different configurations of lexing and parsing by always enabling the local syntax. Using the syntax when the local extension is disabled will still get an error, but from the type-checker.

### DIFF
--- a/parsing/lexer.mll
+++ b/parsing/lexer.mll
@@ -103,9 +103,6 @@ let keyword_table =
 
 let lookup_keyword name =
   match Hashtbl.find keyword_table name with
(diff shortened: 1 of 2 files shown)


stedolan:
This change makes sense but you need to fix the testsuite too (nosyntax.ml in typing-local). Also, there will need to be corresponding ocamlformat and merlin patches, at least.

lpw25:
I think that ocamlformat currently universally enables the extension for its parsing, so IIUC the patch there isn't a prerequisite for releasing this.

lpw25:
Tests updated

PR #20, opened 2022-05-03
Add subtyping between arrows of related modes

lpw25:
Add the obvious subtyping relationship between arrows of related modes, including support in `build_subtype`.

### DIFF
--- a/testsuite/tests/typing-objects/Tests.ml
+++ b/testsuite/tests/typing-objects/Tests.ml
@@ -715,7 +715,7 @@ Error: Multiple definition of the type name t.
 
 fun x -> (x :> < m : 'a -> 'a > as 'a);;
 [%%expect{|
-- : < m : (< m : 'a > as 'b) -> 'b as 'a; .. > -> 'b = <fun>
+- : < m : (< m : 'a -> 'a > as 'a) -> 'a; .. > -> 'a = <fun>
 |}];;
 
 fun x -> (x : int -> bool :> 'a -> 'a);;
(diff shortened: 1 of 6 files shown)


typing/ctype.ml
```diff
@@ -4160,14 +4171,19 @@ let rec build_subtype env visited loops posi level t =
...
+  | Tarrow((l,a,r) , t1, t2, _) ->
       if memq_warn t visited then (t, Unchanged) else
       let visited = t :: visited in
       let (t1', c1) = build_subtype env visited loops (not posi) level t1 in
       let (t2', c2) = build_subtype env visited loops posi level t2 in
-      let c = max c1 c2 in
-      (* FIXME update arrow modes *)
-      if c > Unchanged then (newty (Tarrow(l, t1', t2', Cok)), c)
+      let (a', c3) =
+        if level > 2 then build_submode (not posi) a else a, Unchanged
```
stedolan, line 4180: `level > 2`?

testsuite/tests/typing-objects/Tests.ml
```diff
@@ -715,7 +715,7 @@ Error: Multiple definition of the type name t.
 
 fun x -> (x :> < m : 'a -> 'a > as 'a);;
 [%%expect{|
-- : < m : (< m : 'a > as 'b) -> 'b as 'a; .. > -> 'b = <fun>
+- : < m : (< m : 'a -> 'a > as 'a) -> 'a; .. > -> 'a = <fun>
```
stedolan, line 718: Why did this change?

typing/ctype.ml
lpw25: build_subtype only does its work down to some depth of expansions and object/variant types. It's completely ad-hoc how far to go and after some thought I decided that only replacing mode variables up until the first object/variant after and an expansion was reasonable. That corresponds to `2`.

PR #22, opened 2022-05-06
Add config option to enable/disable stack allocation

lpw25:
The local extension can now be turned on independently of whether anything actually gets allocated on the stack.


lambda/translcore.ml
```diff
@@ -95,20 +94,24 @@ let extract_float = function
...
-       raise (Error (loc, Local_allocs_not_enabled));
-     Alloc_local
+     if Config.stack_allocation then Alloc_local
+     else Alloc_heap
 
-let transl_value_mode loc mode =
+let transl_alloc_mode_assignment alloc_mode =
+  match Btype.Alloc_mode.constrain_lower alloc_mode with
+  | Global -> Assignment
+  | Local -> Local_assignment
```
stedolan, line 107: This should probably be `Assignment` when the extension is disabled

testsuite/tests/typing-local/alloc.ml
```diff
@@ -1,5 +1,11 @@
 (* TEST
-   flags += "-extension local" *)
+   * stack-allocation
+   ** native
+      reference = "${test_source_directory}/alloc.stack.reference"
+   * no-stack-allocation
+   ** native
```
stedolan, line 6: I think this should be no-stack-allocation + no-flat-floatarray, since it has specific alloc counts.

lambda/printlambda.ml
```diff
@@ -603,11 +603,9 @@ let rec lam ppf = function
...
       let pr_params ppf params =
         match kind with
-        | Curried {nlocal} ->
-            let first_local = List.length params - nlocal in
-            List.iteri (fun i (param, k) ->
-                fprintf ppf "@ %a%a%s" Ident.print param value_kind k
-                  (if first_local <= i then "[->L]" else "")) params
+        | Curried _ ->
+            List.iter (fun (param, k) ->
+                fprintf ppf "@ %a%a" Ident.print param value_kind k) params
```
stedolan, line 608: Any particular reason for removing this?

stedolan (changes requested):

stedolan:
As well as `transl_alloc_mode`, local allocations can also be triggered from the `caml_curry` functions in the case `Curried {nlocal=n}` where `n > 0`. I'm not yet sure whether this case can be hit when the extension is disabled (it shouldn't be).

utils/clflags.ml
```diff
@@ -379,6 +379,7 @@ module Extension = struct
   type t = Comprehensions | Local
 
   let all = [ Comprehensions; Local ]
+  let default_extensions = [ Local ]
```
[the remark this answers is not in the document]
lpw25: I think that would make `-disable-all-extensions` always fail

lambda/printlambda.ml
lpw25: It shows up in loads of tests, which would all need to have different versions of their output for the no stack allocation configuration. The information doesn't seem that useful so it seems easier to stop generating it. It also seems slightly better not to have this output in e.g. ocaml-ir where I suspect it will confuse anyone who isn't us.

stedolan:
I think we should also add an `assert Config.stack_allocation` to the `Alloc_local` case in `emit.mlp`: that way, if a local allocation is performed when not configured, we'll get a compiler failure rather than unsoundness.

lpw25:
Yeah, I've added asserts there and similar places. I'll push the changes once I've finished testing them

lpw25:
Pushed now. In addition to the asserts I also made `Lambda.alloc_mode` private with a smart constructor that only creates `Alloc_local` when stack allocation is enabled. I also made `Assignment` use an `alloc_mode` to benefit from the same invariant. This makes the diff much bigger but all the additional changes are trivial.

lambda/translcore.ml
```diff
@@ -1119,7 +1115,7 @@ and transl_function0
       match cases with
       | [] ->
         (* With Camlp4, a pattern matching might be empty *)
-        Alloc_heap, Pgenval
+        alloc_heap, Pgenval
```
stedolan, line 1118: (not this line, but github won't let me comment on a line you didn't change!)

I think the expression `(Tupled, tparams, return, region)` above in `transl_tupled_function` should be preceded by:
```ocaml
    let region = region || not (may_allocate_in_region body) in
```
as in `transl_function0`. This was always a valid optimisation (although we didn't previously bother optimising tupled functions in this way), but with this change it has the added benefit of ensuring that we don't create functions with `region = false` when `Config.stack_allocation` is false.

(I'm not sure whether a `region = false` function can definitely cause problems with this patch, but it seems cleaner to prevent them and this is AFAICT the only place where they can arise)

lambda/lambda.ml
```diff
@@ -46,9 +40,62 @@ type field_read_semantics =
...
+  let alloc_heap = Alloc_heap
+
+  let alloc_local : alloc_mode =
+    if Config.stack_allocation then Alloc_local
+    else Alloc_heap
+
+  let join_mode a b =
+    match a, b with
+    | Alloc_local, _ | _, Alloc_local -> Alloc_local
+    | Alloc_heap, Alloc_heap -> Alloc_heap
```
stedolan, line 58: If you like, I think you can put `join_mode` on the other side of the airtight hatchway by writing it instead as something like:
```ocaml
let join_mode a b =
  match a, b with
  | (Alloc_local as x), _
  | _, (Alloc_local as x) -> x
  | Alloc_heap, (Alloc_heap as x) -> x
```

PR #23, opened 2022-05-23
Fix inclusion checks for primitives

lpw25:
The inclusion check implementation for primitives assumed that moregen forced modes to be equal, which is no longer the case. Replace it with an implementation that just checks both possible types.


stedolan:
Closing in favour of ocaml-flambda/flambda-backend#661

PR #24, opened 2022-06-06
Allow type_argument to weaken return modes of expected function types

stedolan:
If e.g. `fold_left` accepts a local closure, then `(+)` stops being a valid thing to pass to fold_left, as its intermediate return mode is global rather than local. This patch makes `type_argument` introduce fresh mode variables for return modes to allow this case to go through.


lpw25 (approved):
I don't love this. The `type_argument` stuff is already a hack that confuses people, and I don't like that we're only doing it down the return side. However, I assume we can probably ditch this once arity/currying is sorted out, so it seems fine to have for now. Also, whilst `type_argument` is a hack, it was designed for exactly this problem with higher-order functions and multiple arrow types, so we're not making things worse.

stedolan:
> I don't love this. The `type_argument` stuff is already a hack that confuses people, and I don't like that we're only doing it down the return side. However, I assume we can probably ditch this once arity/currying is sorted out, so it seems fine to have for now. Also, whilst `type_argument` is a hack, it was designed for exactly this problem with higher-order functions and multiple arrow types, so we're not making things worse.

My thoughts exactly.

stedolan:
Pushed updated tests.

PR #25, opened 2022-06-21
Propagate escaping_context to Env locks to hint about errors

stedolan:
Adds the following hint when local values are used inside closures that are tail call arguments:
```
Error: The value x is local, so cannot be used inside a closure that might escape
Hint: the closure might escape because it is an argument to a tail call
```
(We already have the hint for values other than closures)

### DIFF
--- a/typing/env.ml
+++ b/typing/env.ml
@@ -214,8 +214,14 @@ module TycompTbl =
 
 type empty = |
 
+type escaping_context =
+  | Return
+  | Tailcall_argument
+  | Tailcall_function
+  | Partial_application
+
 type value_lock =
-  | Lock of { mode : Value_mode.t; }
+  | Lock of { mode : Value_mode.t; escaping_context : escaping_context option }
   | Region_lock
 
 module IdTbl =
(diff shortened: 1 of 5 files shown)


typing/env.ml
```diff
@@ -3277,22 +3283,17 @@ let report_lookup_error _loc env ppf = function
...
+  | Local_value_used_in_closure (lid, context) ->
       fprintf ppf
         "@[The value %a is local, so cannot be used \
            inside a closure that might escape@]"
-        !print_longident lid
-
+        !print_longident lid;
+      begin match context with
+      | Some Tailcall_argument ->
+         fprintf ppf "@.@[Hint: the closure might escape because it \
```
lpw25, line 3293: The `the` should be capitalized to match other hints in the compiler.

typing/typecore.ml
```diff
@@ -224,79 +218,75 @@ let case lhs rhs =
...
 
 let mode_return mode =
-  let position = Tail in
-  let escaping_context = Some Return in
-  let tuple_modes = [] in
-  { position; escaping_context; mode; tuple_modes }
+  { position = Tail;
+    escaping_context = Some Return;
+    mode;
+    tuple_modes = []}
```
lpw25, line 232: I would have rather you'd left these as they were. I prefer the let style because it's much easier to add intermediate definitions later when you need them. It is also more uniform when either all or none of the fields are punned.

lpw25 (approved):
LGTM

typing/env.ml
stedolan: Thanks, fixed.

PR #26, opened 2022-06-21
Improve typing of local unboxed values

stedolan:
This patch makes unboxed type constructors avoid using `mode_subcomponent`, so that they can be e.g. passed to tail calls.

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -3675,7 +3675,6 @@ and type_expect_
       end
   | Pexp_record(lid_sexp_list, opt_sexp) ->
       assert (lid_sexp_list <> []);
-      register_allocation expected_mode;
       let opt_exp =
         match opt_sexp with
           None -> None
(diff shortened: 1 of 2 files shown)


lpw25:
Damn, I made a patch for the same thing: https://github.com/lpw25/ocaml-jst/tree/better-local-for-unboxed-constructors. Seems to be basically the same code, so let's use yours.

typing/typecore.ml
```diff
@@ -5165,13 +5170,21 @@ and type_label_exp create env rmode loc ty_expected
...
-      | Nonlocal -> Value_mode.local_to_regional rmode
-      | Unrestricted -> rmode
+      match label.lbl_global, label.lbl_repres with
+      | Global, _ ->
+         mode_global
+      | Nonlocal, _ ->
+         expected_mode.mode
+         |> Value_mode.regional_to_global
+         |> Value_mode.local_to_regional
+         |> mode_nontail
```
lpw25, line 5180: Would you mind wrapping this up as something like `mode_nonlocal_subcomponent` and put it with all the other `mode_foo` functions.

lpw25 (approved):
LGTM

typing/typecore.ml
lpw25: Actually, I think that's probably unsafe. It would allow something like:
```ocaml
type 'a t = { nonlocal_ escape : 'a } [@@unboxed]

let local_to_nonlocal : local_ 'a -> local_ 'a t =
  fun x -> { escape = x }

let leak x =
  let r = local_ ref x in
  let q = local_to_nonlocal r in
  q.escape
```

lpw25:
This needs to have the issue I mentioned in the comments fixed. After that it is good to merge.

typing/typecore.ml
stedolan: Fixed, the call to `Value_mode.regional_to_global` is back in now.

PR #27, opened 2022-06-22
Build `ocamltest` by default

antalsz:
Changing the version number to `4.12.0+dev-ocaml-jst` (or anything containing the string "+dev", but this seemed suitable) causes `./configure` to build `ocamltest` by default, as well as to turn on `--enable-warn-error`.


stedolan:
Thanks for looking at this! The current behaviour is very annoying.

I don't think this is quite the right fix, though: the contents of VERSION are exposed as `Sys.ocaml_version` which is parsed by various things. Changing it means you'll need to update various tools that expect it to match the version they support, so it's probably less work to change the test in configure.ac instead so that ocamltest gets built regardless of VERSION.

antalsz:
My understanding is that the `VERSION` file is the ground truth for the version of the compiler: if I ask for `Sys.ocaml_version` after building, I get the right answer.

```
% install/bin/ocaml
        OCaml version 4.12.0+dev-ocaml-jst

# Sys.ocaml_version;;
- : string = "4.12.0+dev-ocaml-jst"
```

The definition of `Sys.ocaml_version` is in `stdlib/sys.mlp`, lines 131–133:

```
(* The version string is found in file ../VERSION *)

let ocaml_version = "%%VERSION%%"
```

However, none of the generated files with the actual version number are checked in; I built before submitting the PR so I could check that.

[Judging by the main ocaml/ocaml repo, this changed as of 4.14.0](https://github.com/ocaml/ocaml/blob/4.14.0/VERSION), and the comment in that `VERSION` file tells you how to update the version number, but we still have the older version since we're based on 4.12.

antalsz:
Actually @stedolan was right about this

PR #28, opened 2022-06-23
Allow `make -C testuite promote` to take `TEST` and `LIST` variables

antalsz:
The status quo is that you can run any of

```
make -C testsuite one DIR=tests/typing-modules
make -C testsuite one TEST=tests/typing-modules/printing.ml
make -C testsuite one LIST=file
```

if you want to run tests, but you have to run

```
make -C testsuite promote DIR=tests/typing-modules
```

if you want to promote new test output.  This PR attempts to update the machinery for `promote` to be the same as the machinery for `one`, so you can now say

```
make -C testsuite promote TEST=tests/typing-modules/printing.ml
make -C testsuite promote LIST=file
```

as well.

This also fixes an issue where `promote` didn't correctly depend on (`lib` and?) `tools`.

---

I'm not 100% confident this is implemented correctly, and I don't know if we have the requisite expertise or if we want to kick this upstream – let me know.

### DIFF
--- a/testsuite/Makefile
+++ b/testsuite/Makefile
@@ -101,7 +101,10 @@ default:
 	@echo "  one TEST=f      launch just this single test"
 	@echo "  one DIR=p       launch the tests located in path p"
 	@echo "  one LIST=f      launch the tests listed in f (one per line)"
-	@echo "  promote DIR=p   promote the reference files for the tests in p"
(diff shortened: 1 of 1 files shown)


lpw25 (approved):
Looks roughly right, it's correctness isn't crucial so let's merge it. Upstream can give it a more thorough review.

PR #29, opened 2022-06-24
Make empty types immediate

antalsz:
Updates the logic for determining which types are immediate to include empty types such as `type t = |`.  This will close ocaml/ocaml#9975 once I make the equivalent PR upstream.


lpw25 (approved):
LGTM. Since we don't want to make this a language extension, we should try to upstream it sooner rather than later. Ideally we'd get it in to the next 4.14 release as a bugfix-ish.

PR #31, opened 2022-06-28
Add [@tail] and [@nontail] annotations on applications to control tailcalls

stedolan:
Applications marked [@nontail] are now never tailcalls, even if in tail position, and always preserve backtraces.

Applications marked [@tail] are always tailcalls and must be in tail position. (This is the same as the existing `[@ocaml.tailcall]`, but the syntax is nicer since the annotation is on the application rather than the function being applied)

### DIFF
--- a/lambda/simplif.ml
+++ b/lambda/simplif.ml
@@ -220,12 +220,12 @@ let simplify_exits lam =
   | Lprim(p, ll, loc) -> begin
     let ll = List.map simplif ll in
     match p, ll with
-    (* FIXME does this respect Rc_nontail? *)
         (* Simplify %revapply, for n-ary functions with n > 1 *)
-      | Prevapply Rc_normal, [x; Lapply ap]
-      | Prevapply Rc_normal, [x; Levent (Lapply ap,_)] ->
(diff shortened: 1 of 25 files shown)


lambda/translcore.ml
```diff
@@ -901,6 +902,7 @@ and transl_tupled_cases ~scopes patl_expr_list =
   List.map (fun (patl, guard, expr) -> (patl, transl_guard ~scopes guard expr))
     patl_expr_list
 
+(* FIXME: Does this respect Rc_nontail? *)
```
lpw25, line 905: I think it's safe as it is, but `[@nontail]` will prevent an application from being merged with following application.

lambda/simplif.ml
```diff
@@ -220,6 +220,7 @@ let simplify_exits lam =
   | Lprim(p, ll, loc) -> begin
     let ll = List.map simplif ll in
     match p, ll with
+    (* FIXME does this respect Rc_nontail? *)
```
lpw25, line 223: I think this is safe as it is, but `(x |> foo y) [@nontail]` will compile as `(foo y) x` rather than `foo y x`.

lpw25 (approved):
LGTM

PR #32, opened 2022-06-29
Add the "include_functor" extension.

ccasin:
This extension adds the structure item form "include functor F", where F must be
a functor whose parameter can be "filled in" with the previous contents of the
current module.  This eliminates a common source of boilerplate where modules
create submodules that duplicate some of their contents and exist only for the
purpose of applying a functor.  For example:

```
    module Foo = struct
      module T = struct
        type t = int [@@deriving compare]
      end

      include T
      include Comparable.Make(T)
    end
```

Can now be written:

```
    module Foo = struct
      type t = int [@@deriving compare]

      include functor Comparable.Make
    end
```

Where the only difference is that the latter Foo doesn't have the submodule T.

Limitations:
- Use in the signatures of recursive modules is not supported.  This would require more careful thought about `approx_modtype`.
- Including functors with multiple parameters is not supported.  It's not obvious what could be sensibly done in some cases (e.g., if multiple parameter signatures have elements of the same name but different types).

Informative errors are given in both these cases.


ccasin:
A couple notes for reviewers:

- There's a big diff to `signatures` in `include_functor.ml` that looks scary but is just factoring out and parameterizing two helper functions so that they could be re-used in `include_functor_signatures`.
- In `typemod.ml`, the big diff to `transl_sig` is a) changing it to use accumulators, b) separating the list recursion from the examination of the signature elements, and c) adding code for the include_functor case.  So the diff touches the whole function, but really only the `Psig_include` case has new behavior.  (The other two changes were needed to make available information used by this case).

typing/typemod.ml
```diff
@@ -2446,27 +2518,31 @@ and type_structure ?(toplevel = false) funct_body anchor env sstr =
...
         Tstr_attribute attr, [], env
   in
-  let rec type_struct env sstr =
+  let toplevel_sig = Option.value toplevel ~default:[] in
+  let rec type_struct env sstr str_acc sig_acc =
     match sstr with
-    | [] -> ([], [], env)
+    | [] ->
+      let sig_acc =
+        Misc.take sig_acc (List.length sig_acc - List.length toplevel_sig)
```
lpw25, line 2526: Doing maths on list lengths offends my sensibilities. I think you should just make the recursion have another accumulator, one for using with `include functor` and one for using as the final signature. Alternatively have one accumulator and the toplevel items and then combine them at the point where you find an `include functor`.

testsuite/tests/typing-modules/include_functor.ml
```diff
@@ -0,0 +1,329 @@
...
+  let x : t = 5
+
+  include functor F14
+end;;
+[%%expect{|
+module F14 : functor (X : S) (Y : S) -> sig val z : X.t * Y.t end
+Line 9, characters 18-21:
+9 |   include functor F14
+                      ^^^
+Error: Functors with multiple arguments can not be included directly.
```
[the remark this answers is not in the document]
ccasin: I've changed it as you suggest.  So the old error was:

```
Error: Functors with multiple arguments can not be included directly.
       This functor has type
           functor (X : S) (Y : S) -> sig val z : X.t * Y.t end
       Please apply it to explicit arguments instead.
```

And the new one is:

```
Error: This module is not a functor between structures; it has type
           functor (X : S) (Y : S) -> sig val z : X.t * Y.t end
```

I agree the new one is more uniform, but I'm not entirely convinced it's better.  I can imagine a non expert staring at this error message and thinking "What do you mean it's not between structures?  I see them right there."

Perhaps some kind of hint, like "Hint: To include a functor, it must be applied to all but one of its parameters."?

typing/typemod.ml
ccasin: Yes - all the options seemed bad.  But on reflection, I agree having multiple accumulators is least bad.  I've changed it.

ccasin:
I believe I've addressed all the feedback.  I've also rebased onto `main`.  There was just one small conflict to do with the tailcall support in the extension module.

While working on the fix for the scraping bug, I noticed a few other things:

- We were giving an uninformative error if you tried to include a generative functor.  I've added a clearer error and a test case.
- We weren't handling functors with anonymous parameters.  E.g.:
        
        module type S1 = sig end
        module type S2 = sig end
        
        module type F = functor (_ : S1) -> S2
        
        module type G = sig
          include functor S1
        end
  
  Functors without named parameters are kind of degenerate.  In particular we have no way to do the `nondep_sig` stuff because there's no name for the parameter.  But I think that for the same reason we can just skip it.  So I now allow this and have added a test case.
- While adding the code to scrape the functor's return type, it occurred to me we can probably give a better error than the one we discussed for the multi-parameter functor case.  So for code like:
        
        module F (X : S) (Y : S) = struct
          let z = (X.x, Y.x)
        end
        
        module M = struct
          type t = int
          let x : t = 5
        
          include functor F
        end
  I now give the error:
        
        Line 9, characters 18-19:
        9 |   include functor F
                              ^
        Error: The type of this functor's result is not a signature; it is
               functor (Y : S) -> sig val z : X.t * Y.t end

  To me this looks like the best of the options we've considered.  But I can easily switch back to the one we discussed Friday, if you prefer it.  That would look like:
        
        Error: Expected a module type of the form "functor (X : S1) -> S2".
        This has type
               functor (X : S) (Y : S) -> sig val z : X.t * Y.t end

typing/typemod.ml
```diff
@@ -2519,24 +2516,22 @@ and type_structure ?(toplevel = None) funct_body anchor env sstr =
         Tstr_attribute attr, [], env
   in
   let toplevel_sig = Option.value toplevel ~default:[] in
-  let rec type_struct env sstr str_acc sig_acc =
+  let rec type_struct env sstr str_acc sig_acc sig_acc_with_toplvl =
```
lpw25, line 2537: Maybe there is a better name for this accumulator based on how it is consumed rather than how it is produced?

testsuite/tests/typing-modules/include_functor.ml
```diff
@@ -430,8 +430,192 @@ module F14 : functor (X : S) (Y : S) -> sig val z : X.t * Y.t end
...
+  let equal_t' = X.equal
+end
+
+module M18 = struct
+  type t = int
+  let x = 42
+  let equal = Int.equal
+  include functor F18
+end
+
```
lpw25, line 580: Maybe add a second test showing that you can't include a generative functor inside an applicative functor

lpw25 (approved):
A few minor comments, but this looks good to go

PR #33, opened 2022-07-05
Uniqueness mode

anfelor:


lpw25 (commented):
The test cases look good

typing/printtyped.ml
```diff
@@ -330,11 +330,17 @@ and expression i ppf x =
...
       (i+1) x.exp_extra
   in
-  (match Btype.Value_mode.check_const x.exp_mode with
-  | Some Global -> line i ppf "value_mode global\n"
-  | Some Regional -> line i ppf "value_mode regional\n"
-  | Some Local -> line i ppf "value_mode local\n"
-  | None -> line i ppf "value_mode <modevar>\n");
+  ( let consts = Btype.Value_mode.check_const x.exp_mode in
+    line i ppf "value_mode ";
+    line 0 ppf (match fst consts with
```
anfelor, line 335: This is somewhat ugly as I couldn't find a way to combine two format strings (the way '^' acts on normal strings). Is there a nice operator for that?

typing/typecore.ml
```diff
@@ -2852,6 +2893,10 @@ let is_local_returning_expr e =
   in
   let rec loop e =
     match e.pexp_desc with
+    | Pexp_apply
+        ({ pexp_desc = Pexp_extension({txt = "extension.unique"}, PStr []) },
+         [Nolabel, exp]) ->
+      loop exp
```
lpw25, line 2899: This is a pre-existing issue, but probably these cases should work for `ocaml.unique`, `unique`, `local` and `ocaml.local` too.

typing/typecore.ml
```diff
@@ -2981,6 +3028,10 @@ let rec type_approx env sexp =
         raise(Error(sexp.pexp_loc, env, Expr_type_clash (trace, None, None)))
       end;
       ty2
+  | Pexp_apply
+      ({ pexp_desc = Pexp_extension({txt = "extension.unique"}, PStr []) },
+       [Nolabel, e]) ->
+    type_approx env e
```
lpw25, line 3034: This is a pre-existing issue, but probably these cases should work for `ocaml.unique`, `unique`, `local` and `ocaml.local` too.

typing/typecore.ml
```diff
@@ -5814,6 +5886,9 @@ and type_let
     | Pexp_fun _ | Pexp_function _ -> true
     | Pexp_constraint (e, _)
     | Pexp_newtype (_, e)
+    | Pexp_apply
+      ({ pexp_desc = Pexp_extension({txt = "extension.unique"}, PStr []) },
+        [Nolabel, e])
```
lpw25, line 5891: This is a pre-existing issue, but probably these cases should work for `ocaml.unique`, `unique`, `local` and `ocaml.local` too.

lpw25 (commented):
Looks good so far

parsing/parser.mly
```diff
@@ -2634,33 +2695,52 @@ letop_bindings:
         let and_ = {pbop_op; pbop_pat; pbop_exp; pbop_loc} in
         let_pat, let_exp, and_ :: rev_ands }
 ;
-fun_binding:
-    strict_binding
+fun_binding_gen(wrap):
```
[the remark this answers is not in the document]
anfelor: Flagging this as not implemented as of 0d0c78bc7

typing/typetexp.ml
```diff
@@ -224,12 +232,12 @@ and transl_type_aux env policy mode styp =
       let rec loop acc_mode args =
         match args with
         | (l, arg_mode, arg) :: rest ->
-          let arg_cty = transl_type env policy arg_mode arg in
-          let acc_mode = Alloc_mode.join_const acc_mode arg_mode in
+          let arg_cty = transl_type env policy (fst arg_mode) arg in
+          let acc_mode = Locality_mode.join_const acc_mode (fst arg_mode) in
```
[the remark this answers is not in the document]
anfelor: I have implemented it this way, but I am not sure it is correct. Intuitively, the `acc_mode` should always be unique as each partial application creates a new (unique) allocation? The propagation should then happen for the `FnOnce` flag (and that does depend on whether the function itself is `FnOnce`).

typing/mode.mli
```diff
@@ -0,0 +1,209 @@
...
+(*             Xavier Leroy, projet Cristal, INRIA Rocquencourt           *)
+(*                                                                        *)
+(*   Copyright 1996 Institut National de Recherche en Informatique et     *)
+(*     en Automatique.                                                    *)
+(*                                                                        *)
+(*   All rights reserved.  This file is distributed under the terms of    *)
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
```
anfelor, line 14: Do you have preferences for the copyright @lpw25 ? I have found a few that give credit to Jane Street, should I use one of them instead?

anfelor: This applies to the uniqueness analysis as well

typing/uniquenessAnalysis.ml
```diff
@@ -0,0 +1,557 @@
...
+  | _ -> check_uniqueness_exp_ exp uenv
+
+and check_uniqueness_value_bindings_ vbs uenv =
+  let uenv = List.fold_left
+      (fun uenv vb -> pat_to_map vb.vb_pat (exp_to_parent vb.vb_expr) uenv)
+      uenv vbs in
+  List.fold_left (fun uenv vb -> check_uniqueness_parent vb.vb_expr uenv) uenv vbs
+
+and check_uniqueness_parallel es uenv =
+  let uenv, us = List.fold_left_map (fun u e ->
```
lpw25, line 500: Probably better to stick to using the name `uenv` rather than switching to `u` in a few places. I would bind `orig_last_seen` to `uenv.last_seen` outside this fold and use that when reseting the last_seen. That avoids the need to have multiple `uenv`s in scope at one time.

lpw25 (commented):
Some review for uniquenessAnalysis.ml

typing/uniquenessAnalysis.ml
```diff
@@ -0,0 +1,557 @@
...
+  | Tpat_alias(pat',id, _, mode) ->
+    let uenv = match parent with
+      | NoParent -> uenv
+      | OneParent(p) -> add_mode id mode.uniqueness
+                          (add_mproj p mproj id uenv)
+      | TupleParent(ps, exp) -> mark_tuple_parent_seen id ps exp uenv
+    in pat_to_map_ pat' (OneParent id) None uenv
+  | Tpat_constant(_) -> uenv
+  | Tpat_tuple(ps) -> begin
+      match ps with | [] -> uenv | _ ->
```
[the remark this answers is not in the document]
anfelor: Its not necessary, but convenient for debugging as it prevents adding new projections for singleton constructors with OneParent; this keeps the projection map smaller.

typing/uniquenessAnalysis.ml
```diff
@@ -0,0 +1,557 @@
...
+let is_unique_variable_opt exp_opt =
+  match exp_opt with
+  | Some ({ exp_desc = Texp_ident(p, _, _, _) } as exp') ->
+      begin match ident_option_from_path p with
+      | Some id -> Some (id, has_unique_attr_texp exp')
+      | None -> None
+      end
+  | _ -> None
+
+let rec exp_to_parent exp =
```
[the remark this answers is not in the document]
anfelor: I agree that the logic is similar and have moved them closer together. But I think they can not be unified into one function, because they need to be called separately in `check_uniqueness_value_bindings_`.

typing/typecore.ml
```diff
@@ -4776,18 +4857,23 @@ and type_function ?in_function loc attrs env (expected_mode : expected_mode)
                                           ty_fun,
                                           explanation)))
   in
-  if has_local then
-    eqmode ~loc ~env arg_mode Alloc_mode.local
-      (Param_mode_mismatch ty_expected');
+  if has_local || has_unique then begin
+    let mode = match has_local, has_unique with
+      | true, true -> Mode.Alloc.local_unique
+      | true, false -> Mode.Alloc.local
```
lpw25, line 4863: I think that this case (and the one below) is too restrictive. It should probably have a variable for its uniqueness. That will allow code like the following (maybe put this in the testsuite):

```ocaml
let foo : local_ unique_ string -> unit = fun (local_ s) -> ()
let bar : local_ unique_ string -> unit = fun (unique_ s) -> ()
```

The additional variable (and the `assert false` below) could also be avoided by pulling the equality constraint up into these branches and using more specific equality functions (e.g. using equality on just locality rather than the whole mode).

lpw25 (commented):
I've looked over all of parsing and typing again

typing/mode.ml
```diff
@@ -0,0 +1,668 @@
...
+    { upper = L.max_const;
+      lower = L.min_const;
+      vlower = [];
+      mvid = !next_id;
+      mark = false }
+
+  let newvar () = Amodevar (fresh ())
+
+  let newvar_below = function
+    | Amode c when L.eq_const c L.min_const -> Amode L.min_const, false
```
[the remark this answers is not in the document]
anfelor: No, it is not available in the Lattice `L`, which the solver is parameterized by and also not in the Uniqueness and Locality modules. Should I add it?

typing/typecore.ml
```diff
@@ -3834,15 +3920,16 @@ and type_expect_
       let (record, rmode, label, _) = type_label_access env srecord lid in
       let mode =
         match label.lbl_global with
-        | Global -> Value_mode.global
-        | Nonlocal -> Value_mode.local_to_regional rmode
+        | Global -> Mode.Value.global (* lbl_global==Global also implies Shared *)
+        | Nonlocal -> Mode.Value.local_to_regional rmode
         | Unrestricted -> rmode
       in
+      let mode, _ = Mode.Value.newvar_above mode in
```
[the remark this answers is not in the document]
anfelor: > Maybe we should also only do it if the srecord is an ident or a projection from an ident (or a projection from a projection from ...).

I have not implemented this.

anfelor:
To release this the following items should be completed:

 1. There is currently a bug in `matching.ml` where matches on record fields are bound to a `Llet(Alias, _, _, _, _)`. These let-aliases can be inlined to the place where the bound variable is used if the bound variable is used only once. However, this usage site might come after an overwrite of the record by the `{ unique_ x with ... }` construct. As such, this is unsound and can lead to bugs. Equally, later stages of the compiler might do similar transformations and it should be checked that that is disabled for unique values. To fix the first bug, however, it suffices to make the handling of unique fields in `matching.ml` more similar to that of mutable fields (grep for `lbl_mut`).

 2. With uniqueness analysis the order of evaluation matters. In the spirit of point one, consider `({ unique_ x with foo = ... }, x.foo)`. Thanks to Ocaml's right-to-left evaluation order this is actually sound -- the field `foo` of `x` is accessed before it is overwritten for the first tuple entry. However, we should not codify this evaluation order in the uniqueness analysis. As such, all occurrences of `fold_right` in the current uniqueness analysis should be changed to pass the uenv to a map and later merge.

 3. We currently do not mark borrowed contexts as shared that also contain a unique use of the borrowed variables (e.g. in a separate branch). While this is sound, it can be surprising to users why their code should fail when they delete a branch. See the test `borrow_overwritten_by_once` in `unique.ml` as an example. This can be fixed by constraining the current context as global as soon as we see a borrowed use.

PR #35, opened 2022-07-15
Allow use of `local_` to adjust regions for loops.

ccasin:
This patch changes the local stack region behavior for the body of for loops,
and the body and condition of while loops.  Now, the user may request that no
new stack region is created for these positions, and instead local allocations
happen in the region containing the loop.

This is of limited use in isolation, but is useful with the coming "let mutable"
patch, where we want to allow locals in a loop body to "escape" to mutable
variables declared in the scope containing the loop.

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -3919,38 +3919,51 @@ and type_expect_
         exp_attributes = sexp.pexp_attributes;
         exp_env = env }
   | Pexp_while(scond, sbody) ->
-      let cond =
-        type_expect (Env.add_region_lock env) (mode_var ()) scond
+      let cond_env,wh_cond_region =
+        if is_local_returning_expr scond then env, false
+        else Env.add_region_lock env, true
+      in
+      let wh_cond =
+        type_expect cond_env (mode_var ()) scond
           (mk_expected ~explanation:While_loop_conditional Predef.type_bool)
       in
-      let body =
-        type_statement ~explanation:While_loop_body (Env.add_region_lock env) sbody
+      let body_env,wh_body_region =
+        if is_local_returning_expr sbody then env, false
+        else Env.add_region_lock env, true
+      in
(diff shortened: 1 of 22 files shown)


ccasin:
Likely questions:

Q: Why did you keep the region "flags" on loops at the lambda level?
A: Otherwise, `may_allocate_in_region` (in `translcore.ml`) is potentially quadratic for nested loops.

Q: What is this even good for?
A: Once we add `let mutable`, something like:

```
let mut_test1 () =
  let mutable x = [] in
  for i = 0 to 10 do local_
    x <- (local_ (i :: x))
  done;
  ... use x ...
```  

Will be allowed, with the list cells allocated in the function's region.

typing/typecore.ml
```diff
@@ -3919,38 +3919,51 @@ and type_expect_
         exp_attributes = sexp.pexp_attributes;
         exp_env = env }
   | Pexp_while(scond, sbody) ->
-      let cond =
-        type_expect (Env.add_region_lock env) (mode_var ()) scond
+      let cond_env,wh_cond_region =
+        if is_local_returning_expr scond then env, false
```
stedolan, line 3923: The use of `is_local_returning_expr` means that the following is also a loop without a region:
```ocaml
while cond do
  print_endline "hello";
  incr counter;
  local_ ()
done
```
Is this intentional? It's a bit weird, but consistency with the rule for functions is good. (On the other hand, I'm not a huge fan of the deep inspection in the current rule for functions!)

stedolan (approved):

typing/typecore.ml
ccasin: Yeah, this is weird, but after discussion with Leo I think the conclusion is that consistency mandates it.  In the future it would nice to have different keywords for some of the different uses for the current `local_` keyword.

PR #36, opened 2022-07-20
Add the "let mutable" extension

ccasin:
(Note: This builds on top of the ["local_ for loops" PR](https://github.com/ocaml-flambda/ocaml-jst/pull/35) and should be reviewed after it is merged or by comparing it to that branch)

This extension allows let-bound variables to be marked as mutable, as in:

```
let add_55 y =
  let mutable x = y in
  for i = 1 to 10 do
    x <- x + i
  done;
  x
```

Mutable variables are like "ref"s, but don't give you a way to refer to the ref
cell itself - all uses refer to the stored value.

The typechecker ensures that mutable variables do not escape the scope in which
they are defined.  This static check ensures, roughly, that the variable can
be implemented efficiently as a register, without allocation.  Additionally, it
allows for local values to be stored in mutable variables.


stedolan:
There's a small bug in the handling of the toplevel:
```
# let mutable x = 1;;
# x;;
>> Fatal error: x unbound at toplevel
Exception: Misc.Fatal_error.
```
I think the issue is that the toplevel [extracts the defined names from each phrase](https://github.com/ocaml-flambda/ocaml-jst/blob/main/lambda/translmod.ml#L1609-L1613), and the name `x` shows up in `let_bound_idents` but does not form part of the signature. Perhaps it would be simpler to bad `let mutable` at module level? (`struct let mutable x = 1 end` isn't hugely useful)

I'm also not sure about the choice to make `mutable` part of the pattern: I expected to see `mutable` as an alternative to `rec`, part of the syntax of `let`, but here in the typedtree it is instead part of the syntax of patterns. This doesn't make much sense to me, since it only works at top level, disallowing e.g. `let (mutable x, mutable y) = ...`. Are there plans to someday nest mutable patterns inside other ones?

ccasin:
> I'm also not sure about the choice to make `mutable` part of the pattern: I expected to see `mutable` as an alternative to `rec`, part of the syntax of `let`, but here in the typedtree it is instead part of the syntax of patterns. This doesn't make much sense to me, since it only works at top level, disallowing e.g. `let (mutable x, mutable y) = ...`. Are there plans to someday nest mutable patterns inside other ones?

Yes.  I wrote it the other way first, which led to some ugliness in the typing of lets where a mutability flag needs to be passed to many functions.  Leo suggested this approach of having it with the pattern, but we agree it's also awkward and leads to a lot of ugliness in pattern matching code.  I'll probably make another branch with the old way and compare.

ccasin:
> There's a small bug in the handling of the toplevel:
> 
> ```
> # let mutable x = 1;;
> # x;;
> >> Fatal error: x unbound at toplevel
> Exception: Misc.Fatal_error.

Fun, nice catch!  Will think about the best solution - I agree `let mutable` at the structure level is not really sensible anyway.

stedolan:
> Yes - this is unpleasant. I wrote it the other way first, which led to some ugliness in the typing of lets where a mutability flag needs to be passed to many functions. Leo suggested this approach of having it with the pattern, but we agree it's also awkward and leads to a lot of ugliness in pattern matching code. I'll probably make another branch with the old way and compare.

Another suggestion: I think the only valid patterns for `let mutable` are `Ppat_var` and `Ppat_constraint(Ppat_var,...)` (that is, `let mutable x = ...` and `let mutable x : t = ...`). It might be simpler to recognise those two and generate a new `Texp_letmutable` typedtree node instead of plumbing mutability through the pattern matching typechecker.

mshinwell:

PR #37, opened 2022-07-21
Simpler symbols

lukemaurer:
Some cleanups for `Symbol`, `Compilation_unit`, and `Compilenv` on the way to the next PR, which will further streamline the handling of symbols by having each global `Ident.t` point directly to its `Compilation_unit.t`.


middle_end/closure/closure.ml
```diff
@@ -63,7 +63,16 @@ let rec build_closure_env env_param pos = function
    contain the right names if the -for-pack option is active. *)
 
 let getglobal dbg id =
-  Uprim(P.Pread_symbol (Compilenv.symbol_for_global id), [], dbg)
+  assert (Ident.is_global_or_predef id);
```
lukemaurer, line 66: This code really doesn't belong here - almost exactly the same code needs to be added to `Opttoploop` for the same purpose: translating a global `Ident.t` to a linkage name. `Compilenv.symbol_for_global` used to do this, and `Compilenv` still needs to be consulted anyway, so we should just keep `Compilenv.symbol_for_global` after all (with this new implementation).

middle_end/closure/closure.ml
```diff
@@ -63,7 +63,16 @@ let rec build_closure_env env_param pos = function
...
+  let pack_prefix =
+    if Ident.is_global id then Compilenv.pack_prefix_for_global_ident id
+    else Compilation_unit.Prefix.empty
+  in
+  let symbol =
+    Symbol.for_ident id ~pack_prefix
+    |> Symbol.linkage_name
+  in
```
lukemaurer, line 74: This code really doesn't belong here - almost exactly the same code needs to be added to `Opttoploop` for the same purpose: translating a global `Ident.t` to a linkage name. `Compilenv.symbol_for_global` used to do this, and `Compilenv` still needs to be consulted anyway, so we should just keep `Compilenv.symbol_for_global` after all (with this new implementation).

mshinwell: I'm not sure about this (one thing is that Flambda 1 doesn't seem to need such an interface in `Compilenv`).  When `Ident.t` knows the `Compilation_unit.t`, `Compilenv` won't need to be involved, so it may be better to tolerate the code duplication in this intermediate state of the work.

lpw25 (commented):
Some comments from when we looked over the code yesterday

utils/compilation_unit.mli
```diff
@@ -0,0 +1,144 @@
...
+   do different things w.r.t. the prefix. *)
+
+(** A distinguished compilation unit for initialisation of mutable state. *)
+val dummy : t
+
+(** A distinguished compilation unit for predefined exceptions. *)
+val predef_exn : t
+
+(** The name of the compilation unit, excluding any [for_pack_prefix]. *)
+val name : t -> Name.t
```
lukemaurer, line 104: Perhaps this should be `basename` by analogy to Unix. (This is particularly desirable since a `Compilation_unit.t` is (currently) often called a "name", and asking the basename of a name is perfectly sensible but asking the name of a name is not.)

utils/compilation_unit.mli
```diff
@@ -0,0 +1,144 @@
...
+      as if it was generated with "-for-pack". *)
+  val extract_prefix : string -> t * Name.t
+
+  val to_list_outermost_pack_first : t -> Name.t list
+
+  val to_string : t -> string
+
+  val is_empty : t -> bool
+
+  val first_component_to_name : t -> Name.t option
```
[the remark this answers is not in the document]
lukemaurer: Turns out it's actually unused (and anyway matching on `to_list` is almost as easy). Got rid of it.

utils/compilation_unit.mli
```diff
@@ -0,0 +1,144 @@
...
+type error = private
+  | Invalid_character of char
+  | Bad_compilation_unit_name of string
+
+(** The exception raised by conversion functions in this module. *)
+exception Error of error
+
+(** The type used for recording digests of compilation units in .cmo and
+    .cmx files. *)
+type crcs = (t * Digest.t option) list
```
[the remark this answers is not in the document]
lukemaurer: Deleted. (Once we get to `.cmi`s, this type _could_ replace `Misc.crcs`, but even then I'm not sure it's needed: are the modules in a `crcs` list ever not just the base names of modules in the same pack - meaning we don't need to store the prefix?)

middle_end/closure/closure.ml
lukemaurer: I'd still rather it be in just one place, even if we're about to get rid of it. It doesn't need to be in `Compilenv` necessarily.

asmcomp/cmm_helpers.ml
```diff
@@ -2725,6 +2725,18 @@ let emit_float_array_constant symb fields cont =
...
+  let compilation_unit =
+    match compilation_unit with
+    | None -> Compilation_unit.get_current_exn ()
+    | Some compilation_unit -> compilation_unit
+  in
+  let compilation_unit_symbol =
+    Symbol.for_compilation_unit compilation_unit
+    |> Symbol.linkage_name
+  in
+  compilation_unit_symbol ^ Symbol.separator ^ name
```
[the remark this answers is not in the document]
lukemaurer: I rewrote this to leave the string manipulation to `Symbol`. `Cmm_helpers.make_symbol` is still there for the convenient optional argument and call to `Symbol.linkage_name`, which seems much more "helper"-like.

lukemaurer:
Closing because this ended up as ocaml-flambda/flambda-backend#753.

PR #40, opened 2022-08-08
Weaken guarantees for tail calls

stedolan:
OCaml currently guarantees that any call at the end of a function body is a tail call, allowing tail-recursive loops to be computed in constant space.

However, this guarantee applies whenever a call is the last thing a function does, even when no tail recursion is intended. This has unfortunate implications for local allocations. For instance:
```ocaml
let print_all s xs =
  List.iter (fun x -> Printf.printf "%s: %d\n" s x) xs

let iter_list f xs =
  let rec loop = function
    | [] -> ()
    | x :: xs -> f x; loop xs
  in
  loop xs
```
In the `print_all` example, we would like to use local allocations to stack-allocate the closure passed to `List.iter`. However, since this is a tail call, we cannot: tail calls must free any stack allocations *before* the call, since cleaning them up afterwards would break the constant-space tail call property.

There is no particular reason for this call to List.iter to be a tail call: it just happens to be the last thing this function does. The asymptotic space complexity of this program cannot be affected by whether or not tail call optimisation applies here, and disabling it allows the closure to be stack-allocated.

In the `iter_list` example, the `loop` function must remain tail-recursive as the space complexity of the program depends upon it. However, the call to loop at the very end of `iter_list` need not be a tail call, and allowing it not to be means that the closure for `loop` itself may be stack-allocated.

---

The problem is that it is not visible in the source whether a call in tail position is an intentional tail call or just happens to be at the end of a function. This patch introduces a heuristic for distinguishing them. The heuristic does not always choose correctly, but can be overridden with explicit `[@tail]` or `[@nontail]` annotations.

Rather than choosing a very complicated heuristic which guesses correctly as often as possible, the idea here is to choose a simple, understandable heuristic, but to also implement a complicated warning which warns about cases where the simple heuristic might not work and asks the programmer for an explicit annotation.

## Tailcall heuristic

The simple heuristic proposed here is:

> **Calls to previously-defined functions are not tailcalls (unless annotated)**

A "previously-defined function" means a functions defined at module level or locally whose definition is complete: in particular, inside the body of `let rec f () = ...`, `f` is not "previously-defined", so normal tail recursion continues to work. Functions that are passed as parameters are never "previously-defined".

This heuristic allows local allocations in the two examples above, while correctly ensuring that there are tailcalls in e.g. the following examples:

```ocaml
let rec simple_recursion n =
  if n <= 0 then 0 else simple_recursion (n-1)

let rec map_cps f l k =
  match l with
  | [] -> k []
  | x :: xs -> map_cps f xs (fun t -> k (f x :: t))
```

It gets the wrong answer (removing a tailcall, worsening the asymptotic space complexity) in e.g. the following examples:
```ocaml
let rec outer_loop a b =
  let inner x = outer_loop a x in
  if a+b <= 0 then 0 else inner (b-1)

let (>>=) = Option.bind
let rec bindy (b : string option) =
  b >>= fun b ->
  int_of_string_opt b >>= fun b ->
  if b < 10 then bindy (Printf.sprintf "%s0" b) else b
```
In `outer_loop`, the function `inner` counts as "previously-defined" and is not tail-called, while the same happens in `bindy` for the function `>>=`. (See "Warning" below for more on these examples)

### Alternative heuristics

Two heuristics that were (eventually!) rejected are:
 
**Only calls to recursive bindings are tailcalls (unless annotated)**: This is an even simpler rule, making all calls non-tail except annotated and direct ones (e.g. in `simple_recursion`). This rule would require a large amount of annotation for backwards-compatibility. In particular, functions like `map_cps` would, if not annotated, stop running in constant stack space. This is a problem, as this pattern is used specifically to ensure constant stack space usage.

**Only calls in tail position of a let-rec function are tailcalls (unless annotated)**: This is also simpler than the rule proposed here, but gets the wrong answer more frequently. (In particular, function calls at the end of the base case of recursive functions become tail, even if they are calls to e.g. List.fold_left or some other function that need never be tailcalled)

## Warning

As well as the heuristic described above, there is a warning which detects certain calls in tail position which are not tail calls but look like maybe they should be. The current implementation of the warning warns about calls to a previously-defined local function `f` if:

  - `f` contains a tail call to an recursively-bound identifier
  - all calls to `f` are in tail position

This catches the `outer_loop` example above (although not `bindy`). There are a number of annotations needed to the compiler codebase to prevent this warning occurring, added in the most recent commit.

## TODO

  - [ ] There is commented-out extended warning logic meant to detect cases like `bindy`. It currently has too many false positives, so should be fixed or deleted.
  - [ ] There should be some way to revert to the old behaviour for all uses of some function (e.g. Option.bind)
  - [ ] There should be some way to revert to the old behaviour for an entire module / file.
  - [ ] This change should probably be under a `-extension` flag


mshinwell:

PR #42, opened 2022-08-10
Ensure that (if c then E) counts E as tail position for typing

stedolan:
This fixes a bug in the local allocations typechecker's handling of tail calls, where if expressions with no `else` part were not recognised as being in tail position.


lpw25 (approved):
This fixes the bug. It could probably have a slightly more relaxed mode for the body of the `if`, but it would make little difference in practice.

PR #43, opened 2022-08-15
Add [@tail hint] annotation to select default behaviour explicitly

stedolan:
This patch adds `[@tail_if_possible]`, which explicitly selects the current default tailcall behaviour (which may change with #40), but is not an error if used outside of tail position.

There is no reason for a programmer to ever write this annotation, since they're better off using `[@tail]` or `[@nontail]`. However, ppx generators such as `ppx_let` that may rely on tail-call behaviour can use this annotation to guarantee the current behaviour is preserved, without having to work out whether a given application is in tail position.

### DIFF
--- a/parsing/builtin_attributes.ml
+++ b/parsing/builtin_attributes.ml
@@ -94,6 +94,11 @@ let mk_internal ?(loc= !default_loc) name payload =
   Attr.mk ~loc name payload
 
 
+let ident_of_payload = function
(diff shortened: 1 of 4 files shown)


lpw25:
Maybe something like `[@tail hint]` would be better? Saves us updating ppxlib and similar with another compiler attribute.

stedolan:
> Maybe something like [@tail hint] would be better? Saves us updating ppxlib and similar with another compiler attribute.

Good idea, changed.

lpw25 (approved):

PR #47, opened 2022-09-19
Immutable arrays

antalsz:
This patch introduces immutable arrays.  It builds on top of the comprehensions PR (#46); in particular, it builds on top of the `Extensions` machinery I introduced there and improves it.  I may get around to writing up a long PR description here, but this is it for now.

A couple thoughts, not related to each other:

1. We/the middle end need to add a more optimizable `get` operation for immutable arrays.  This can be part of a separate PR.
2. The following syntax is available for something: `(|1; 2; 3; 4|)`.  I considered using it for immutable arrays but it didn't have an obvious `get` operator (we currently have `[:1; 2; 3; 4:].:(0)`).  Still, might be nice?


ceastlund:
Name bikeshedding: Have we considered anything longer than "iarray"? I like short names, but I'm not sure the "i" tells much of a story on its own, and "Iarray.create" just sort of looks wrong to me. I'll suggest "immarray", although I'm aware that "imm" could be "immutable" or "immediate" as well.

ceastlund:
I'll also add that `Immarray` has pleasing symmetry with `Dynarray` from an upstream ocaml PR (https://github.com/ocaml/ocaml/pull/11563)

riaqn:
Just out of curiosity: I feel this would play nicely with the data race freedom thing? You will create an immutable array which can be in-place updated if you exclusively have it; and even strongly update it if it is unique!

PR #48, opened 2022-09-28
Layouts prototype

ccasin:
This is a work-in-progress version of layouts.  The goal here is to support inference for a very simple system of just 5 layouts: `any`, `value`, `immediate`, `immediate64` and `void`.   We don't have `#`, and are extremely conservative about where we allow non-value layouts.

There are still many known bugs and plenty of stuff left to do.  But some simple examples are starting to work - see `testsuite/tests/typing-layouts/basics.ml`.

There are many `XXX`s in the code.  Some of these have been resolved but not cleaned up yet.  Many are just places I wanted to come back and double check.  Others are known bugs.


typing/ctype.mli
```diff
@@ -398,14 +406,33 @@ val get_current_level: unit -> int
...
+val is_void_type : Env.t -> type_expr -> bool
 
 (* Stubs *)
 val package_subtype :
     (Env.t -> Path.t -> Longident.t list -> type_expr list ->
       Path.t -> Longident.t list -> type_expr list -> bool) ref
 
 val mcomp : Env.t -> type_expr -> type_expr -> unit
+
+val get_unboxed_type_representation : Env.t -> type_expr -> type_expr
```
goldfirere, line 420: We may need to update our terminology. This function might look through `[@@unboxed]` things, or it might somehow get the layout of something. The fact that we use "unboxed" to refer to two notionally separate concepts is going to cause trouble, forever.

Inlined types?
Raw types?
Unpackaged types?

Maybe I'm overreacting, but I foresee many, many conversations that involve "Which 'unboxed' do you mean there?".

typing/ctype.mli
```diff
@@ -398,14 +406,33 @@ val get_current_level: unit -> int
...
+
+(* Cheap upper bound on layout.  Will not expand unboxed types - call
+   [type_layout] if that's needed. *)
+val estimate_type_layout : Env.t ->  type_expr -> Type_layout.t
+val type_layout : Env.t -> type_expr -> Type_layout.t
+val type_sort : Env.t -> type_expr -> sort
```
goldfirere, line 426: What if it doesn't have a sort? Does this throw an exception?

typing/outcometree.mli
```diff
@@ -126,7 +133,7 @@ and out_type_decl =
     otype_params: out_type_param list;
     otype_type: out_type;
     otype_private: Asttypes.private_flag;
-    otype_immediate: Type_immediacy.t;
+    otype_layout: out_layout_annot option;
```
goldfirere, line 136: At a glance, this doesn't seem equivalent to the old bit of information here. This new piece sounds like it's about surface syntax, whereas the old one sounds like it's about something computed during type-checking. But I could be wrong, as I didn't look deeply.

typing/typedtree.mli
```diff
@@ -250,12 +250,13 @@ and expression_desc =
       expression * Longident.t loc * Types.label_description * expression
   | Texp_array of expression list
   | Texp_ifthenelse of expression * expression * expression option
-  | Texp_sequence of expression * expression
+  | Texp_sequence of expression * Types.layout * expression
```
goldfirere, line 253: This surprises me. Why is it necessary? If it's just a cached value, please add a comment saying so.

typing/typedtree.mli
```diff
@@ -250,12 +250,13 @@ and expression_desc =
...
   | Texp_while of {
       wh_cond : expression;
       wh_cond_region : bool; (* False means allocates in outer region *)
       wh_body : expression;
-      wh_body_region : bool  (* False means allocates in outer region *)
+      wh_body_region : bool;  (* False means allocates in outer region *)
+      wh_body_layout : Types.layout
```
goldfirere, line 259: This is very similar to the layout in `Texp_sequence`. I remain curious why it's necessary. (I'll stop writing this on other similar fields.)

typing/typedtree.mli
```diff
@@ -487,6 +489,7 @@ and value_binding =
   {
     vb_pat: pattern;
     vb_expr: expression;
+    vb_sort: Types.sort;
```
goldfirere, line 492: What if a value is immediate?

utils/misc.mli
```diff
@@ -487,7 +487,12 @@ module Bitmap : sig
 end
 
 module EnvLazy: sig
-  type ('a,'b) t
+  type ('a,'b) t = ('a,'b) eval ref
```
goldfirere, line 490: This concretization is surprising. Maybe a debugging fragment?

goldfirere (commented):
As promised, I took a look through the mli files (only). Looks very reasonable. I ask a few clarifying questions, but it's sensible to delay acting on these until after the debugging is complete.

ccasin:
@goldfirere  Thanks. I've pushed fixes for a few of the nits, and will respond to some of the other comments inline.

typing/types.mli
```diff
@@ -377,23 +392,31 @@ type type_declaration =
...
+                       runtime_tag: int}    (* The runtime tag *)
+        | Extension of Path.t
+
 and record_representation =
-    Record_regular                      (* All fields are boxed / tagged *)
-  | Record_float                        (* All fields are floats *)
-  | Record_unboxed of bool    (* Unboxed single-field record, inlined or not *)
-  | Record_inlined of int               (* Inlined record *)
-  | Record_extension of Path.t          (* Inlined record under extension *)
+  | Record_unboxed of layout
```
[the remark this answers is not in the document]
ccasin: I've added a comment for the one that might be a little confusing (but I think the old code errs in the other direction - surely "Inlined record" is an unhelpful comment for `Record_inlined of int` :) )

typing/ctype.mli
ccasin: I'd be interested to know how much `[@@unboxed]` is used.  If it's quite uncommon, maybe we don't need to panic.

(Note that this code isn't actually added by my PR, it's just showing up in the diff because is moved in upstream PR #10041, which has been pulled in here).

typing/typedtree.mli
ccasin: We have layout annotations in the type tree when the layout of a subexpression isn't obvious from the expression.  This is needed to compile `void` (and maybe other layouts in some cases?) because we want to add a handler for the static exception the void thing will raise.

Here and in the loop cases, I'm allowing void things in statement positions.

typing/typedtree.mli
ccasin: The `sort` is sufficient to know how to compile the pattern match (and in fact, after discussion with Leo, I plan to change the other layouts in the typed tree to sorts, defaulted to value for `any` cases, as that's enough information).

typing/typedtree.mli
goldfirere: That makes sense, then. But I expect it will can be removed once we re-do the implementation of `void`. Could you maybe leave a CR-someday to revisit later?

typing/typedtree.mli
goldfirere: Hm. Interesting. That makes sense. So, for any situation that can actually be compiled (that is, not `any`), all we need is the sort. That sounds right. Still, we do certain optimizations over immediates, and so storing only the sort -- while sufficient -- might miss these optimizations. My very uneducated guess is that the immediate-optimizations (which pre-date layouts, of course) look elsewhere. Maybe that's all OK, but I wonder if there's an opportunity for consolidation. In any case, I agree that `sort` works here.

typing/typedtree.mli
ccasin: Indeed - I meant to say I plan to add a comment at the top explaining what the layouts are doing and what might change in the future.

PR #49, opened 2022-10-05
Support building with dune

stedolan:
This patch allows the compiler to be built using an existing installation of OCaml 4.12 and dune, which has two advantages:
  - it's faster than the current build system
  - Merlin works

To use it, you'll need OCaml and dune (and preferably merlin) preinstalled, e.g. via opam. Then, after running `configure`, build and run the testsuite with:

    make -f Makefile.jst runtest-upstream

Other useful targets include `compiler` (to build without running the tests) and `hacking` (to run a polling dune build).

The existing build system (`make world`, etc.) still works, but if you use it you need to `make clean` before trying the dune-based build, as dune does not like build artifacts in the source tree.

Currently, the `ocamldoc` tests are broken with this setup, so the tests won't pass unless you configure with `--disable-ocamldoc`. I've updated github actions to use the new build system, but actions is broken right now so this is untested.


stedolan:
After far too many rounds of single combat with Github Actions, this seems to actually build (even ocamldoc tests, even in 32-bit configurations).

> May I suggest adding a note to e.g. HACKING.adoc? 

Good idea. I added a new HACKING.jst.adoc to avoid merge conflicts with upstream.

PR #50, opened 2022-10-06
Add support for `global_` and `nonlocal_` constructor arguments

riaqn:
This patch adds the support for `global_` and `nonlocal_` annotations to constructor arguments, such that upon pattern matching the arguments extracted will be considered as global regardless of the mode of the construct, or nonlocal in relation to the mode of the construct. Those invariants will be checked when constructing, of course.


typing/includecore.ml
```diff
@@ -375,8 +383,11 @@ let rec compare_constructor_arguments ~loc env params1 params2 arg1 arg2 =
         Some (Arity : constructor_mismatch)
       else if
         (* Ctype.equal must be called on all arguments at once, cf. PR#7378 *)
-        Ctype.equal env true (params1 @ arg1) (params2 @ arg2)
-      then None else Some Type
+        Ctype.equal env true (params1 @ List.map fst arg1) (params2 @ List.map fst arg2)
+      then
+        (* now check global flags *)
+        Option.map (fun err -> Nonlocality err) (List.find_map (fun (x,y) -> compare_global_flags x y) (List.combine (List.map snd arg1) (List.map snd arg2)))
```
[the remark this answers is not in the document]
lpw25: It would be clearer if the steps were separated out:
```ocaml
  then begin
    let gfs = List.combine arg_gfs1 arg_gfs2 in
    let err = List.find_map (fun (gf1, gf2) -> compare_global_flags gf1 gf2) gfs in
    Option.map (fun e -> Nonlocality e) err
  end
  ```
  or if you prefer:
  ```ocaml
  then
     List.combine arg_gfs1 arg_gfs2
     |> List.find_map (fun (gf1, gf2) -> compare_global_flags gf1 gf2)
     |> Option.map (fun e -> Nonlocality e)
  ```

File: typing/typedecl.ml [hunk 2]
```diff
@@ -249,11 +251,27 @@ let transl_labels env closed lbls =
...
+  in
+  let tyl_gfl = List.map mk tyl in
+  let tyl_gfl' =
+    List.map
+      (fun (cty,gf) ->
+         let ty = cty.ctyp_type in
+         (ty, gf)
+      )
+      tyl_gfl in
+  tyl_gfl, tyl_gfl'
```
lpw25, line 268: Style is a bit hard to read here. More usual would be:
```ocaml
  let tyl_gfl' =
    List.map
      (fun (cty, gf) ->
         let ty = cty.ctyp_type in
         ty, gf)
      tyl_gfl
  in
```
and you can probably shorten that to:
```ocaml
  let tyl_gfl' = List.map (fun (cty, gf) -> cty.ctyp_type, gf) tyl_gfl in
```

typing/oprint.ml
```diff
@@ -443,6 +444,7 @@ and print_out_label ppf (name, mut_or_gbl, arg) =
   in
   fprintf ppf "@[<2>%s%s :@ %a@];" flag name print_out_type arg
 
+
```
lpw25, line 447: No need for extra blank line

typing/oprint.ml
```diff
@@ -724,12 +726,12 @@ and print_out_type_decl kwd ppf td =
         print_private td.otype_private
         print_record_decl lbls
   | Otyp_sum constrs ->
-      let variants fmt constrs =
+    let variants fmt (constrs : (string * (out_type * out_global) list * out_type option) list) =
```
lpw25, line 729: Probably clearer without the type annotation.

typing/oprint.ml
```diff
@@ -745,29 +747,55 @@ and print_out_type_decl kwd ppf td =
     print_immediate
     print_unboxed
 
-and print_out_constr ppf (name, tyl,ret_type_opt) =
+
+and print_simple_out_gf_type ppf (ty, gf) =
+  let en = Clflags.Extension.is_enabled Local in
```
lpw25, line 752: `en` is a bit short for this. `locals_enabled` probably better.

typing/oprint.ml
```diff
@@ -745,29 +747,55 @@ and print_out_type_decl kwd ppf td =
...
+  let (name, tyl, ret_type_opt) = a in
   let name =
     match name with
     | "::" -> "(::)"   (* #7200 *)
     | s -> s
   in
   match ret_type_opt with
   | None ->
-      begin match tyl with
+      begin match (tyl : (out_type * out_global) list) with
```
lpw25, line 785: Probably clearer without the type annotation.

testsuite/tests/typing-local/nosyntax.ml
```diff
@@ -30,3 +30,33 @@ let local_ref (f : lfn -> unit) =
 [%%expect{|
 val local_ref : (lfn -> unit) -> unit = <fun>
 |}]
+
+
+type foo = Foo of string
+type gfoo = GFoo of (string [@ocaml.global])
+(* native global_ syntax is not supported yet *)
```
lpw25, line 37: I think this parses now, so we should test this case.

testsuite/tests/typing-local/nosyntax.ml
```diff
@@ -30,3 +30,33 @@ let local_ref (f : lfn -> unit) =
...
+Error: This value escapes its region
+|}]
+
+
+let cast2 ((x : gfoo) [@ocaml.local]) = match x with
+  GFoo s -> Foo s
+[%%expect{|
+val cast2 : (gfoo [@local]) -> foo = <fun>
+|}]
+
```
lpw25, line 60: I'm not sure why these tests are here. Instead I would have expected to see some tests which print `[@global]` in error messages e.g. errors for incorrect module inclusion around global fields.

testsuite/tests/typing-local/local.ml
```diff
@@ -2253,3 +2253,246 @@ module F :
...
     sig val foo : (float -> string) bi end
 |}]
+
+
+(*
+ * constructor arguments global
+ *)
+
+(* Global argument are preserved in module inclusion *)
+(* TODO: what about non-local? *)
```
lpw25, line 2263: You should address this TODO

testsuite/tests/typing-local/local.ml
```diff
@@ -2253,3 +2253,246 @@ module F :
...
+         sig type t = Bar of global_ string end
+       Type declarations do not match:
+         type t = Bar of string
+       is not included in
+         type t = Bar of global_ string
+       Constructors do not match:
+         Bar of string
+       is not compatible with:
+         Bar of global_ string
+       The second is global and the first is not.
```
lpw25, line 2287: This error message is a bit confusing. It says that the constructor is global where as it is the argument to the constructor that is global. This would be even more confusing with multiple arguments. Something along the lines of:
```
The second has a global argument and the first does not
```
would be good for the single argument case. Whilst something like:
```
The second has a global argument at position 3 and the first does not.
```
would be needed when there are more arguments.

testsuite/tests/typing-local/local.ml
```diff
@@ -2253,3 +2253,246 @@ module F :
...
+
+(* but global is outer than local *)
+let f (s : string) =  (* s is global *)
+  let local_ x = RFoo (s, "bar") in  (* x is local *)
+  "foo"
+[%%expect{|
+Line 2, characters 13-14:
+2 |   let local_ x = RFoo (s, "bar") in  (* x is local *)
+                 ^
+Warning 26 [unused-var]: unused variable x.
```
lpw25, line 2430: You should probably use `_` instead of `x` to avoid this warning

lpw25 (commented):
Looks correct. I've suggested various small changes.

typing/ctype.ml
```diff
@@ -2491,6 +2491,19 @@ and mcomp_variant_description type_pairs env xs ys =
   in
   iter xs ys
 
+and mcomp_tuple_description type_pairs env =
+  let rec iter x y =
+    match x, y with
+    | (ty1, gf1) :: xs, (ty2, gf2) :: ys ->
+      mcomp type_pairs env ty1 ty2;
+      if gf1 == gf2
```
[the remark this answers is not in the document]
riaqn: Yes my bad - I just habitually wrote `==` for equality.

typing/includecore.ml
```diff
@@ -375,8 +383,11 @@ let rec compare_constructor_arguments ~loc env params1 params2 arg1 arg2 =
         Some (Arity : constructor_mismatch)
       else if
         (* Ctype.equal must be called on all arguments at once, cf. PR#7378 *)
-        Ctype.equal env true (params1 @ arg1) (params2 @ arg2)
-      then None else Some Type
+        Ctype.equal env true (params1 @ List.map fst arg1) (params2 @ List.map fst arg2)
```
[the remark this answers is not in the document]
riaqn: Yes that's more efficient.

testsuite/tests/typing-local/nosyntax.ml
riaqn: Currently both record fields and constructor arguments are broken in this regards when local extension disabled: they would accept `global_` syntax without error; and record fields in particular also prints `global_` instead of `[@global]`. I suggest that I make another patch fixing record fields, then base this patch on that.

testsuite/tests/typing-local/local.ml
riaqn: `let local_ _ = ...` would curiously trigger an error pointing to the underscore, saying 'Syntax Error`

testsuite/tests/typing-local/local.ml
riaqn: Yes thanks for bringing this up.

riaqn:
the latest commit fixed most issues except the two unresolved ones.

riaqn:
The latest commit fixed issues that occurs when the local extension is disabled. In this case, `global_` and `nonlocal_` annotating record fields and constructor arguments should trigger error, and only `[@global]` and `[@nonlocal]` should be accepted and printed.

riaqn:
I have rebased the PR on top of the 4.14 update.

typing/typedecl.ml
```diff
@@ -204,7 +206,32 @@ let make_params env params =
     with Already_bound ->
       raise(Error(sty.ptyp_loc, Repeated_parameter))
   in
-    List.map make_param params
+  List.map make_param params
```
lpw25, line 209: Unnecessary blank line

typing/typedecl.ml
```diff
@@ -204,7 +206,32 @@ let make_params env params =
...
+      else
+        raise(Error(loc, Global_and_nonlocal))
+    end
+  else
+    begin
+      if not nonlocal then
+        Types.Unrestricted
+      else
+        Types.Nonlocal
+    end
```
lpw25, line 234: This set of `if`s probably reads more easily as:
```ocaml
match global, nonlocal with
| true, true -> raise(Error(loc, Global_and_nonlocal))
| true, false -> Types.Global
| false, true -> Types.Nonlocal
| false, false -> Types.Unrestricted
```

typing/typedecl.ml
```diff
@@ -204,7 +206,32 @@ let make_params env params =
...
+
+
+let transl_global_flags loc attrs =
+  let transl_global_flag loc (r : (bool,unit) result) =
+    match r with
+    | Ok b -> b
+    | Error () -> raise(Error(loc, Local_not_enabled))
+  in
+  let global = transl_global_flag loc (Builtin_attributes.has_global attrs)
+  and nonlocal = transl_global_flag loc (Builtin_attributes.has_nonlocal attrs)
```
lpw25, line 219: It is better to only use `and` for recursive definitions.

PR #51, opened 2022-10-14
add `[%exclave]` support

riaqn:
This patch adds the support for the `[%exclave]` extension, which can be applied to an expression and results in early ending of the current region before the expression is evaluated. As such, it can only appear in the tail position of the region. A motivating example is as follows:
```
let foo (local_ x) = 
let local_ y = (complex computation on x) in 
if y then [%exclave] (local_ None)
else  [%exclave] (local_ (Some x))
```
where `foo` is supposed to be some helper function that takes arguments from parent region and returns values also in parent region. The intermediate results (`y`) it uses are stored in its local stack and released upon entering `[%exclave]`. As a result, you are not allowed to refer to `y` in `[%exclave]`.

### DIFF
--- a/jane/doc/local-reference.md
+++ b/jane/doc/local-reference.md
@@ -512,9 +512,9 @@ count + 1}` inside `[%exclave]` to release the region before the allocation.
 After the revision, the function should use O(1) stack space.
 
 
-`[%exclave]` terminates the current region, so values within it cannot be used
-inside `[%exclave]`. For example, the following code produces an error because
-`x` would escape its region:
+`[%exclave]` terminates the current region, so local values from that region
+cannot be used inside `[%exclave]`. For example, the following code produces an
+error because `x` would escape its region:
 
 
 ```
--- a/lambda/simplif.ml
+++ b/lambda/simplif.ml
@@ -355,7 +355,7 @@ let simplify_exits lam =
   | Lifused(v, l) -> Lifused (v,simplif ~layout ~try_depth l)
   | Lregion (l, ly) -> Lregion (
       simplif ~layout ~try_depth:(try_depth + 1) l,
-      result_layout ly) (* lwhite suggest to double check this case *)
+      result_layout ly)
   | Lexclave l -> Lexclave (simplif ~layout ~try_depth:(try_depth - 1) l)
   in
   simplif ~layout:None ~try_depth:0 lam
@@ -957,17 +957,19 @@ let simplify_local_functions lam =
     with_scope ~scope:lam lam
   and region lam =
     let old_tail_scope = !current_region_scope in
+    let old_scope = !current_scope in
     current_region_scope := Some !current_scope;
     current_scope := lam;
     tail lam;
-    current_scope := Option.get !current_region_scope;
+    current_scope := old_scope;
     current_region_scope := old_tail_scope
   and exclave lam =
     let old_current_scope = !current_scope in
+    let old_tail_scope = !current_region_scope in
     current_scope := Option.get !current_region_scope;
     current_region_scope := None;
     tail lam;
-    current_region_scope := Some !current_scope;
+    current_region_scope := old_tail_scope;
     current_scope := old_current_scope
   and function_definition lf =
     let old_function_scope = !current_function_scope in
--- a/testsuite/tests/typing-local/loop_regions.ml
+++ b/testsuite/tests/typing-local/loop_regions.ml
@@ -111,7 +111,6 @@ let loc_func () =
         r := local_stack_offset ()
     ) in
   let offset1 = local_stack_offset () in
-
   fun_exclave offset_func;
   let offset2 = local_stack_offset () in
   [offset1; !offset_func; offset2]
@@ -125,7 +124,6 @@ let nonloc_func () =
     r := local_stack_offset ()
   in
   let offset1 = local_stack_offset () in
-
   fun_nonexclave offset_func;
   let offset2 = local_stack_offset () in
   [offset1; !offset_func; offset2]
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -4673,16 +4673,14 @@ and type_expect_
         exp_attributes = sexp.pexp_attributes;
         exp_env = env }
   | Pexp_while(scond, sbody) ->
-      let cond_env,mode=
-        Env.add_region_lock env, mode_region Value_mode.local
-      in
+      let cond_env = Env.add_region_lock env in
+      let mode = mode_region Value_mode.local in
       let wh_cond =
         type_expect cond_env mode scond
           (mk_expected ~explanation:While_loop_conditional Predef.type_bool)
       in
-      let body_env,position =
-        Env.add_region_lock env, RTail (Value_mode.local, Nontail)
-      in
+      let body_env = Env.add_region_lock env in
+      let position = RTail (Value_mode.local, Nontail) in
       let wh_body =
         type_statement ~explanation:While_loop_body
           ~position body_env sbody
@@ -4706,9 +4704,8 @@ and type_expect_
       let for_id, new_env =
         type_for_loop_index ~loc ~env ~param
       in
-      let new_env, position =
-        Env.add_region_lock new_env, RTail (Value_mode.local, Nontail)
-      in
+      let new_env = Env.add_region_lock new_env in
+      let position = RTail (Value_mode.local, Nontail) in
       let for_body =
         type_statement ~explanation:For_loop_body ~position new_env sbody
       in
--- a/testsuite/tests/typing-local/regions.ml
+++ b/testsuite/tests/typing-local/regions.ml
@@ -29,11 +29,11 @@ let[@inline never] uses_exclave x =
   let local_ r = ref x in
   let _ = opaque_identity r in
   [%exclave] (
-    ()
+    check_empty "cleanup upon exclave"
   )
 let () =
-  uses_local 42;
-  check_empty "cleanup upon exclave"
+  uses_exclave 42
+
 
 let[@inline never] uses_local_try x =
   try


testsuite/tests/typing-local/unregion.ml
```diff
@@ -0,0 +1,74 @@
+(* TEST
+ * expect *)
+
+(* TYPING TESTS *)
+(* TYPING TESTS *)
+(* TYPING TESTS *)
+(* the following two tests show that y is regional *)
```
lpw25, line 7: This comment is a little confusing. It isn't really that `y` is regional, it's that it is local for the caller's region rather than this function's region. I'd probably reword this to something like "the following two tests show that y can leave the current function despite being local".

testsuite/tests/typing-local/unregion.ml
```diff
@@ -0,0 +1,74 @@
...
+5 |     ref y
+            ^
+Error: This value escapes its region
+|}]
+
+let foo x =
+  [%unregion] (
+    let local_ y = Some x in
+    (* y is not local either *)
+    (* because it can escape *)
```
lpw25, line 25: This comment is a bit confusing. `y` is local it's just part of the caller's region. It would probably be better to say that `y` can escape the function even though it is local.

testsuite/tests/typing-local/unregion.ml
```diff
@@ -0,0 +1,74 @@
...
+[%%expect{|
+Line 3, characters 10-30:
+3 |   let z = [%unregion] (Some y) in
+              ^^^^^^^^^^^^^^^^^^^^
+Error: Unregion expression should only be in tail position
+|}]
+
+(* SEMANTICS TESTS *)
+(* SEMANTICS TESTS *)
+(* SEMANTICS TESTS *)
```
lpw25, line 58: Duplicated comment.

lpw25 (commented):
A few changes needed but mostly looks good.

lambda/simplif.ml
```diff
@@ -911,7 +925,7 @@ let simplify_local_functions lam =
         let curr_scope =
           match ap_region_close with
           | Rc_normal | Rc_nontail -> !current_scope
-          | Rc_close_at_apply -> !current_region_scope
+          | Rc_close_at_apply -> Option.get !current_region_scope  (* will raise if None *)
```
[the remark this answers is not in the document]
riaqn: Yes should we have a dedicated exception for this?

riaqn:
This PR has been rebased on top of version 4.14.

riaqn:
The last commit made it so that any function ending with `[%unregion]` will always return `local_`. This is to accommodate some code in the compiler which relies on that bit to decide if the function might allocate in parent region. This also gives the expected output for a test case that wasn't working before (and I didn't know why).

antalsz:
We should strongly consider switching to using `[@unregion]` (separately, I want to make that change for `[%local]`); in terms of making sure code is compatible with standard OCaml, `unregion` is perfectly ignorable.  Also, if we want to support an `unregion_` keyword, then we also need to update our syntax-rewriter to support `[@unregion]`; that's a job for me (as long as the location information is propagated correctly), just let me know when it's necessary.

goldfirere:
Yes -- let's use attributes for anything that can safely be just ignored in the upstream compiler.

riaqn:
Hmm do you think I should make the change in this PR, or do it some time later?

goldfirere:
Antal may see things differently, but I think it's best to merge patches when they're not going to cause trouble for others. I *think* the infrastructure to handle e.g. `[@unregion]` correctly is still some time away, and so merging what you have and then quickly (this week?) changing it sounds reasonable to me.

antalsz:
Because we can handle `[@unregion]` correctly via the refactor-based rewriter, I'd prefer making that change in ocaml-jst before we roll it out to users.  Whether that means in this PR or a different PR is something I'm happy to leave in your hands, @riaqn.  Also, are we planning to add syntax for it?

typing/typecore.ml
```diff
@@ -374,7 +393,12 @@ let mode_argument ~funct ~index ~position ~partial_app alloc_mode =
                 Id_prim _), 1, Tail ->
      (* The second argument to (&&) and (||) is in
         tail position if the call is *)
-     mode_return (Value_mode.local_to_regional vmode)
+      (* CR zqian: this is suspicious, 
+         because vmode is not the mode of the region;
+         ideally the `position` should contain the mode 
+         of the region so we can pass it to mode_return;
+         but this might be fine because of mode-crossing? *)
+     mode_return vmode
```
riaqn, line 401: This one needs some inspection.

riaqn:
This PR relies on https://github.com/ocaml-flambda/flambda-backend/pull/1131 to be merged safely.

typing/typecore.ml
```diff
@@ -380,7 +407,12 @@ let mode_argument ~funct ~index ~position ~partial_app alloc_mode =
                 Id_prim _), 1, Tail ->
      (* The second argument to (&&) and (||) is in
         tail position if the call is *)
-     mode_return (Value_mode.local_to_regional vmode)
+      (* CR zqian: this is suspicious,
+         because vmode is not the mode of the region;
+         ideally the `position` should contain the mode
+         of the region so we can pass it to mode_return;
+         but this might be fine because of mode-crossing? *)
```
lpw25, line 414: I agree this is dubious. I wonder if we can treat `%sequor` and `%sequand` differently elsewhere rather than here. Maybe handling them where we handle `%ignore` in `type_application` would be easier.

typing/typecore.ml
```diff
@@ -4143,6 +4177,38 @@ and type_expect_
...
+            With the introduction of explicit region we will need something more clever *)
+          let mode' = mode_default mode in
+          (* enforce to be local; part of the compiler depends on this bit to know
+          if it allocate in parent region. *)
+          submode ~loc ~env ~reason:Other Value_mode.local mode';
+          let env' = Env.add_exclave_lock env in
+          let exp =
+            type_expect ?in_function ~recarg env' mode' sbody ty_expected_explained
+          in
+          (* this whole thing returns regional value *)
```
lpw25, line 4202: This comment doesn't add anything

typing/typecore.ml
```diff
@@ -4143,6 +4177,38 @@ and type_expect_
...
+          raise (Error (loc, env, Exclave_in_nontail_position))
+        | RTail (mode, _) ->
+          (* mode' is non-tail-of-region, because currently our language cannot construct
+           region in the tail of another region.
+            With the introduction of explicit region we will need something more clever *)
+          let mode' = mode_default mode in
+          (* enforce to be local; part of the compiler depends on this bit to know
+          if it allocate in parent region. *)
+          submode ~loc ~env ~reason:Other Value_mode.local mode';
+          let env' = Env.add_exclave_lock env in
```
lpw25, line 4198: `new_env` is the usual name for this

typing/typecore.ml
```diff
@@ -4143,6 +4177,38 @@ and type_expect_
...
+          let exp =
+            type_expect ?in_function ~recarg env' mode' sbody ty_expected_explained
+          in
+          (* this whole thing returns regional value *)
+          submode ~loc ~env ~reason:Other Value_mode.regional expected_mode;
+          { exp_desc = Texp_exclave exp;
+            exp_loc = loc;
+            exp_extra = [];
+            exp_type = exp.exp_type;
+            exp_env = env;  (* use the old env *)
```
lpw25, line 4208: Not worth a comment

typing/typecore.ml
```diff
@@ -4143,6 +4177,38 @@ and type_expect_
...
+      if (txt = "extension.exclave") && not (Language_extension.is_enabled Local) then
+          raise (Typetexp.Error (loc, Env.empty, Unsupported_extension Local));
+      begin
+        match expected_mode.position with
+        | RNontail ->
+          raise (Error (loc, env, Exclave_in_nontail_position))
+        | RTail (mode, _) ->
+          (* mode' is non-tail-of-region, because currently our language cannot construct
+           region in the tail of another region.
+            With the introduction of explicit region we will need something more clever *)
```
lpw25, line 4193: I would probably just say `RNontail` rather than "non-tail-of-region". The second line has weird indentation. s/construct region/construct a region/. The second sentence doesn't really add anything.

typing/typecore.ml
```diff
@@ -4143,6 +4177,38 @@ and type_expect_
...
+          let mode' = mode_default mode in
+          (* enforce to be local; part of the compiler depends on this bit to know
+          if it allocate in parent region. *)
```
lpw25, line 4196: The wording here isn't very good. Perhaps:
> The middle-end relies on all functions which allocate into their parent's region having a return mode of local.

typing/typecore.ml
```diff
@@ -4613,20 +4679,21 @@ and type_expect_
         exp_attributes = sexp.pexp_attributes;
         exp_env = env }
   | Pexp_while(scond, sbody) ->
-      let cond_env,wh_cond_region =
-        if is_local_returning_expr scond then env, false
-        else Env.add_region_lock env, true
+      let cond_env,wh_cond_region,mode=
+      (* not checking is_local_returning_expr because you should just use exclave *)
```
lpw25, line 4683: The comment only makes sense in relation to the diff. Best to remove it.

typing/typecore.ml
```diff
@@ -4613,20 +4679,21 @@ and type_expect_
         exp_attributes = sexp.pexp_attributes;
         exp_env = env }
   | Pexp_while(scond, sbody) ->
-      let cond_env,wh_cond_region =
-        if is_local_returning_expr scond then env, false
-        else Env.add_region_lock env, true
+      let cond_env,wh_cond_region,mode=
```
lpw25, line 4682: With the `if` gone, this code can now be tidied up. For instance `wh_cond_region` is no longer needed since it is always just `true`.

typing/typecore.ml
```diff
@@ -4613,20 +4679,21 @@ and type_expect_
...
       in
       let wh_cond =
-        type_expect cond_env mode_local scond
+        type_expect cond_env mode scond
           (mk_expected ~explanation:While_loop_conditional Predef.type_bool)
       in
-      let body_env,wh_body_region =
-        if is_local_returning_expr sbody then env, false
-        else Env.add_region_lock env, true
+      (* local because unit cross modes  *)
```
lpw25, line 4690: I think this comment is confusing. I would say that it's local because the value isn't going to escape the region. Probably just remove the comment.

typing/typecore.ml
```diff
@@ -4613,20 +4679,21 @@ and type_expect_
...
+      let body_env,wh_body_region,position =
```
lpw25, line 4691: This can be refactored now

typing/typecore.ml
```diff
@@ -4637,22 +4704,22 @@ and type_expect_
...
       in
       let for_id, new_env =
         type_for_loop_index ~loc ~env ~param
       in
-      let new_env, for_region =
-        if is_local_returning_expr sbody then new_env, false
-        else Env.add_region_lock new_env, true
+      let new_env, for_region, position =
+      (* local because unit cross modes  *)
+        Env.add_region_lock new_env, true, RTail (Value_mode.local, Nontail)
```
lpw25, line 4719: Same as above

lpw25 (commented):
Looks pretty good. Should be good to go once various superficial things are fixed.

lambda/simplif.ml
```diff
@@ -352,7 +355,8 @@ let simplify_exits lam =
   | Lifused(v, l) -> Lifused (v,simplif ~layout ~try_depth l)
   | Lregion (l, ly) -> Lregion (
       simplif ~layout ~try_depth:(try_depth + 1) l,
-      result_layout ly)
+      result_layout ly) (* lwhite suggest to double check this case *)
```
lpw25, line 358: Comment no longer needed

testsuite/tests/typing-local/loop_regions.ml
```diff
@@ -98,12 +101,43 @@ let nonloc_while_cond () =
...
+let loc_func () =
+  let offset_func = ref (-1) in
+  let fun_exclave r =
+    [%exclave] (
+        let z = local_ (Some (Sys.opaque_identity 42)) in
+        let _ = (opaque_local z) in
+        r := local_stack_offset ()
+    ) in
+  let offset1 = local_stack_offset () in
+
```
lpw25, line 114: Weird blank line

testsuite/tests/typing-local/loop_regions.ml
```diff
@@ -98,12 +101,43 @@ let nonloc_while_cond () =
...
+let nonloc_func () =
+  let offset_func = ref (-1) in
+  let fun_nonexclave r =
+    let z = local_ (Some (Sys.opaque_identity 42)) in
+    let _ = (opaque_local z) in
+    r := local_stack_offset ()
+  in
+  let offset1 = local_stack_offset () in
+
```
lpw25, line 128: Weird blank line

typing/typecore.ml
```diff
@@ -4679,25 +4673,23 @@ and type_expect_
         exp_attributes = sexp.pexp_attributes;
         exp_env = env }
   | Pexp_while(scond, sbody) ->
-      let cond_env,wh_cond_region,mode=
-      (* not checking is_local_returning_expr because you should just use exclave *)
-        Env.add_region_lock env, true, mode_region (Value_mode.local)
+      let cond_env,mode=
+        Env.add_region_lock env, mode_region Value_mode.local
```
lpw25, line 4677: No need for this to be a single `let` with a tuple, just have two `let`s

typing/typecore.ml
```diff
@@ -4679,25 +4673,23 @@ and type_expect_
...
       in
       let wh_cond =
         type_expect cond_env mode scond
           (mk_expected ~explanation:While_loop_conditional Predef.type_bool)
       in
-      (* local because unit cross modes  *)
-      let body_env,wh_body_region,position =
-        Env.add_region_lock env, true, RTail (Value_mode.local, Nontail)
+      let body_env,position =
+        Env.add_region_lock env, RTail (Value_mode.local, Nontail)
```
lpw25, line 4684: Same as above

typing/typecore.ml
```diff
@@ -4714,16 +4706,15 @@ and type_expect_
       let for_id, new_env =
         type_for_loop_index ~loc ~env ~param
       in
-      let new_env, for_region, position =
-      (* local because unit cross modes  *)
-        Env.add_region_lock new_env, true, RTail (Value_mode.local, Nontail)
+      let new_env, position =
+        Env.add_region_lock new_env, RTail (Value_mode.local, Nontail)
```
lpw25, line 4710: And this one

jane/doc/local-reference.md
```diff
@@ -442,39 +442,81 @@ be the caller's (or that of the caller's caller, etc., if the caller is also
...
+  | [] -> Ok { res = 0 }
+  | x :: xs ->
+      match p x with
+      | false -> None
+      | true ->
+          match count_result f xs with
+          | None as r -> r
+          | Some { res = count } ->
+              Some { res = count + 1 }
+```
```
lpw25, line 497: Let's just take this example out for now. I'll add a better one in a later PR.

jane/doc/local-reference.md
```diff
@@ -442,39 +442,81 @@ be the caller's (or that of the caller's caller, etc., if the caller is also
...
+in which case the result of the entire function would be `None`. It is designed
+not to allocate heap memory, instead using the stack for all `Some` allocations.
+However, it will currently use O(N) stack space because all allocations occur in
+the original caller's stack frame. To improve its space usage, we remove the
+`local_` annotation (so the function has its own region), and wrap `Some {res =
+count + 1}` inside `[%exclave]` to release the region before the allocation.
+After the revision, the function should use O(1) stack space.
+
+
+`[%exclave]` terminates the current region, so values within it cannot be used
```
lpw25, line 515:
```suggestion
`[%exclave]` terminates the current region, so local values from that region cannot be used
```

lpw25 (approved):
Small small suggestions then this is good to go

jane/doc/local-reference.md
riaqn: is it better to have a silly but working example, than to have no example?

PR #52, opened 2022-10-17
Add support for partial application

riaqn:
This patch allows users to use underscore (`_`) as argument place holders to denote parial applications. This is useful due to the poor interactions between `local_` and implicit partial application.

Currently it is implemented as simple syntax sugar; `foo _ x` simply unfolds into 
```
fun y -> [%unregion] (foo y x)
```
where `[%unregion]` is introduced in https://github.com/ocaml-flambda/ocaml-jst/pull/51 and is intended to remove the default region from the eta-expansion function, to avoid `local_` related problems.


riaqn:
closed in favor of https://github.com/ocaml-flambda/flambda-backend/pull/1888

PR #53, opened 2022-10-20
Replace var with local for faster mode checking

riaqn:
During type checking there were several places where the returning value's mode can be relaxed to `local`, thus making mode checking faster.

### DIFF
--- a/testsuite/tests/formatting/test_locations.dlocations.ocamlc.reference
+++ b/testsuite/tests/formatting/test_locations.dlocations.ocamlc.reference
@@ -118,7 +118,7 @@ let rec fib = function | 0 | 1 -> 1 | n -> (fib (n - 1)) + (fib (n - 2))
                     expression (test_locations.ml[19,572+9]..test_locations.ml[19,572+20])
                       Texp_apply
                       apply_mode Default
-                      alloc_mode <modevar>
+                      alloc_mode global
                       expression (test_locations.ml[19,572+9]..test_locations.ml[19,572+12])
                         Texp_ident "fib"
                       [
@@ -127,7 +127,7 @@ let rec fib = function | 0 | 1 -> 1 | n -> (fib (n - 1)) + (fib (n - 2))
                           expression (test_locations.ml[19,572+13]..test_locations.ml[19,572+20])
                             Texp_apply
                             apply_mode Default
-                            alloc_mode <modevar>
+                            alloc_mode global
                             expression (test_locations.ml[19,572+16]..test_locations.ml[19,572+17])
                               Texp_ident "Stdlib!.-"
                             [
@@ -146,7 +146,7 @@ let rec fib = function | 0 | 1 -> 1 | n -> (fib (n - 1)) + (fib (n - 2))
                     expression (test_locations.ml[19,572+23]..test_locations.ml[19,572+34])
                       Texp_apply
                       apply_mode Default
-                      alloc_mode <modevar>
+                      alloc_mode global
                       expression (test_locations.ml[19,572+23]..test_locations.ml[19,572+26])
                         Texp_ident "fib"
                       [
@@ -155,7 +155,7 @@ let rec fib = function | 0 | 1 -> 1 | n -> (fib (n - 1)) + (fib (n - 2))
                           expression (test_locations.ml[19,572+27]..test_locations.ml[19,572+34])
                             Texp_apply
                             apply_mode Default
-                            alloc_mode <modevar>
+                            alloc_mode global
                             expression (test_locations.ml[19,572+30]..test_locations.ml[19,572+31])
                               Texp_ident "Stdlib!.-"
                             [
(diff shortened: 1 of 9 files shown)


lpw25 (approved):
LGTM

lpw25:
The failing test suggests this might somehow be interacting with inlining. We should investigate.

lpw25:
We should be able to revive this now that `exp_mode` is gone.

Note that it does still require a small adjustment to the typing of function application. Currently `Texp_apply` gets the expected mode, it should instead get the more accurate mode cacluated in `type_application`. So `type_application` should be modified to return it's `mode` value and that should be used instead.

stedolan:
For this to work, we need submoding to be admissable: whenever `G |- e : A @ global`, then also `G |- e : A @ local`. That way, typechecking against a fresh variable mode is no more powerful than typechecking against the constant mode `local`. I know of two places where this property current fails:

  - As Leo points out, the expected mode is copied directly to the alloc_mode of `Texp_apply`. Instead this should be the return mode of the applied function (which is always below the expected mode, but need not be equal)
  - In `type_function`, `alloc_mode` is defined to be equal to the expected mode. For most data structures this is fine (since alloc mode is only constrained above by expected mode), but for functions the alloc mode is additionally constrained above by the return mode (for uncurried functions). To resolve this, the alloc_mode for functions should be a fresh variable below the expected mode, at least when `uncurried_function = true`

As a separate PR, we should also make the same change for mode-crossing - it introduces lots of fresh variables that could be the constant modes local or global.

lpw25:
> To resolve this, the alloc_mode for functions should be a fresh variable below the expected mode, at least when uncurried_function = true

We should also take are to ensure that we still guarantee that `local_ (fun x y -> ...)` is guaranteed not to allocate, but adding something to the expected mode type to indicate that we have been explicitly asked for it to be local.

However, I thought that the `type_function` issue couldn't affect this PR because it only sets expected modes to local when the type is `unit`, `int` or `bool`, and so a function literal cannot be involved.

riaqn:
Some notes for myself:
If `expected_mode` is the upper bound, not the exact mode of the expression being type checked, then there are some other places where we need to insert mode variables. For example, `Texp_construct` and `Texp_record` both contains an `alloc_mode` directly corresponding to `expected_mode`. If `expected_mode` is `local` which means "no restriction', the `alloc_mode` of those allocations will actually just be `local` instead of undeterminited. Howerver, that is fine because we will push them upwards anyway in `optimise_allocations`.

riaqn:
> When we do that we should also take are to ensure that we still guarantee that local_ (fun x y -> ...) is guaranteed not to allocate, by adding something to the expected mode type to indicate that we have been explicitly asked for it to be local.

Are we talking about the `alloc_mode` of functions? I thought they will be pushed upwards? (Not guarantees, though)

riaqn:
> In type_function, alloc_mode is defined to be equal to the expected mode. For most data structures this is fine (since alloc mode is only constrained above by expected mode), but for functions the alloc mode is additionally constrained above by the return mode (for uncurried functions). To resolve this, the alloc_mode for functions should be a fresh variable below the expected mode, at least when uncurried_function = true

This all makes sense, but will require more work than I thought. The issue is that in the case of uncurried function, we require `alloc_mode < ret_mode`, and this `ret_mode` is later used to type the inner function. Now if `alloc_mode` is `local`, that would force `ret_mode` to be `local`, which is good. We then type check the inner function with `expected_mode = local` which only says "no constraint" instead of "must be local".

To solve this thoroughly, I second @lpw25 's proposal: we should add another field `mode_lower` to `expected_mode` that represents the lower bound (while `expected_mode.mode` still represents the higher bound). The aforementioned `ret_mode` should be the `expected_mode.mode_lower` when type checking the inner function.

lpw25:
> We then type check the inner function with expected_mode = local which only says "no constraint" instead of "must be local".

I think you can just make sure that if the `expected_mode` is "must be local" (by which I assume you mean it has been annotated with `local_`) on an uncurried function then it passes that on to the `expected_mode` that it gives to `type_cases` for the next function.

I really think that we should not try and change the meaning of `expected_mode.mode` -- it's current meaning is really the only one that makes sense.

riaqn:
> I think you can just make sure that if the expected_mode is "must be local" (by which I assume you mean it has been annotated with local_) on an uncurried function then it passes that on to the expected_mode that it gives to type_cases for the next function.

Do you mean to use an extra field in `expected_mode`? Sure that can work too.

> I really think that we should not try and change the meaning of expected_mode.mode -- it's current meaning is really the only one that makes sense.

Could you give an example where "expected_mode means the exact mode of the expression being checked" might cause issues? That's quite suprising to me because previously `exp_mode` is exacty `expected_mode.mode` (which, granted, causes some bugs; but I'm not sure if those bugs are related to this?)

lpw25:
> Do you mean to use an extra field in expected_mode?

Yes

> Could you give an example where "expected_mode means the exact mode of the expression being checked" might cause issues?

It would require making a fresh mode variable for almost every subexpression. Consider:
```ocaml
let foo (local x) =
  if p then x else 5
```
The subexpressions `x` and `5` would need to have separate "expected mode"s because the exact mode for `5` is `global` whilst `x` must be `local. Fundamentally, the expected mode means an upper bound: any expectation or requirement you might make of an expression's mode must just be an upper bound on that mode due to submoding.

riaqn:
This PR is now ready for review. Failing tests are only because I didn't update the expected output for some backends, and will be fixed easily. @stedolan @lpw25 

- replaces var with local for faster mode checking (including in mode crossing)

- lock in env is now alloc_mode

- add exact to expected_mode so now type_function is sensible

lpw25 (approved):
A couple of small suggestions but otherwise looks good

PR #55, opened 2022-10-26
OCaml 4.14

stedolan:
(This PR needs to be merged with a merge commit, not squash or rebase)

Merge TODO:

  - [x] merge & commit conflicts
  - [x] merge CHANGES, VERSION, dotfiles, etc.
  - [x] resolve all added/removed files (only conflicts left)
  - [x] main Makefiles: Makefile.compilerlibs, stdlib/Makefile, stdlib/StdlibModules
  - [x] configure, make clean, dune b some trivial targets
  - [x] regenerate parser & build parsing/ utils/ via Dune
  - [x] build typing/
  - [x] build lambda/
  - [x] build ocamlc
  - [x] build runtime & test some bytecode
  - [x] build stdlib
  - [x] build ocamlopt
  - [x] run 'make -f Makefile.jst compiler'
  - [x] run testsuite

Things to check / recheck / fix:

  - [x] Restore boot/ocamlc, boot/ocamllex and .depend files
  - [x] Decide what should go in VERSION and update build-aux/OCAML__VERSION_EXTRA
  - [x] fix upstream build system
  - [x] reapply $(git show 4bbde7a0f0c196bb5f83d5ee0baa224c89323745 -- toplevel/opttoploop.ml) to toploop.ml
  - [x] ~~restore ocaml-variants.opam (?)~~
  - [x] possibly restore 4ac4cfd4b831f6a5ec8668c31c24d22d4733fae0 or check
  - [x] verify that stdlib module capitalisation is OK
  - [x] recheck error handling in Ctype: Unify should not be raised, Moregen etc. should be caught. (equal, moregen)
  - [x] recheck filter_arrow, filter_method API change.
  - [x] ~~Percent_free_def: 100 vs 120 - probably 120 is better?~~
  - [x] Pexp_send and Leo's "Represent ancestor variables more directly" - check that this still works ("selfpat-" case)
  - [x] Recheck lazy subst performance (with original benchmarks)
  - [x] Check modes and Curried and alloc_heap in tmc.ml.
  - [x] split_default_wrapper: check that Pisint triggers
  - [x] ensure that build_apply respects max_arity: #10719
  - [x] Fix probes in amd64/emit.mlp (env handling trickiness)
  - [x] Bug in nongen: odoc_parameter is mli-less and triggers local error
  - [x] make ocamldebug work
  - [x] fix runtime/sak error (why is this being generated in the source tree?)
  - [x] ref_spec.compilers.reference: var_rec = (makemutable 0 (int,*) 0 65), set_open_poly [int]
  - [x] testsuite in flat-float-array mode
  - [x] 32-bit builds
  - [x] apply typecore scope change (backtrack in loosen_ret_modes)
  - [x] run testsuite with debug runtime (port flambda-backend fixes)


lambda/tmc.ml
```diff
@@ -557,7 +558,11 @@ let llets lk vk bindings body =
   ) bindings body
 
 let find_candidate = function
-  | Lfunction lfun when lfun.attr.tmc_candidate -> Some lfun
+  | Lfunction lfun when lfun.attr.tmc_candidate ->
+     if not lfun.region then
+       raise (Error (Debuginfo.Scoped_location.to_location lfun.loc,
+                     Tmc_without_region));
+     Some lfun
```
lpw25, line 565: This part makes me a little nervous. Using `lfun.region` as a measure of whether something might allocate in the parent region is dubious. Tail-calls can generally break this and `unregion` will also allow breaking this. I'd much prefer that we add a constraint at type-checking time on functions marked for trmc.

lpw25 (approved):
I've reviewed the comments that aren't resolving merge conflicts.

lambda/tmc.ml
lpw25: Tails-calls to functions without `lfun.region` themselves will still allocate in the parent region

stedolan: I don't see how that is possible, can you construct an example? Surely you get `region=false` if you tail call a function that has `region=false`, otherwise this seems broken.

lpw25: As discussed: this is fine but we should probably rename this field to make it's meaning clearer, and we'll need to adjust the unregion patch to keep it accurate.

PR #58, opened 2022-11-01
Allow immediates to cross modes

riaqn:
This patch allows immediates (such as integers) to cross modes. That means a locally "allocated" integer can be safely coreced into a heap "allocated" integer, because an integer is immediately available without referring to anything on the stack or heap. 

For performance considerations, we only check mode crossing at several specific points which should cover most use cases. Some cases remains the old way as reflected in the tests, which might be fixed in the future.


stedolan (changes requested):
This looks good to me!

I'm trying to distill down a bit the set of situations where mode crossing occurs. So far, I think it's:

  1. When checking against a principally-known type `t`, mode-cross if `t` is immediate
  2. When inferring a principally-known type `t`, mode-cross if `t` is immediate
  3. When adding a principally-known type `t` to the environment, mode-cross if `t` is immediate

In theory, (1) could be done at every `type_expect`, but to avoid doing it redundantly (which might be slow), the patch only does it when a new expected type arises (e.g. arguments of function application, record construction, explicit annotation, which all happen to hit `type_argument`).

Likewise, case (2) happens when we learn a new inferred type, at e.g. record projection, or results of function application).

Case (3) does mode-crossing when types are entered into the environment, rather than during typechecking of variable references, for IIUC two reasons:

  - Context locking: in `let x : int = ... in fun y -> ... x ...`, we want to do the mode-crossing before `x` is added to the environment, so that it's OK to close over it in a closure that might escape.
  - Performance: it's better to do the check on variable definition rather than use, because a variable can be used many times.

Does that characterisation of when to mode cross make sense to you? If so, I have two questions / suggestions:

  - Should we also mode-cross in introduction forms of builtin types (that is, `Pexp_tuple`, `Pexp_array`, `Pexp_lazy`), as these are places where a new expected type arises? For instance, it might be nice to support the following, which I believe will not currently typecheck:
    ```
    let f (local_ x) : int * int = (x, x)
    ```

  - What's the rationale for the exact places that mode crossing occurs in patterns? By my understanding of how this works, it makes sense to me to do mode crossing on variable patterns only, but it's done on a different set of patterns here. (From the tests it looks like it hits more or less the right set of cases, but I'm curious how the code was chosen here)

testsuite/tests/typing-local/local.ml
```diff
@@ -1903,24 +1903,6 @@ let primloc x = let local_ y = Int32.add x 1l in Int32.to_int y
...
-[%%expect{|
-val testbool1 : bool -> bool = <fun>
-Line 7, characters 20-21:
-7 |   true && (false || b)
-                        ^
-Error: This local value escapes its region
-  Hint: Cannot return local value without an explicit "local_" annotation
-|}]
-
-
```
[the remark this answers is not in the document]
riaqn: these tests no longer fail because of mode crossing. Instead of deleting them, we can find some other tests that shows that the two operators tail call on the right.

ceastlund:
Question - what are the implications of this for cases where an abstract type might be represented as an immediate, but we still might want to track whether references to it can escape some scope?

Case 1: Integer-encoded pointers into the C heap, where use of `local_` might make it easier to know when to call `free`.

Case 2: Indexes into an interned string table, and a function `intern_temporarily : string -> f:(local_ t -> 'a) -> 'a` that can reclaim entries.

lpw25:
That's not supported in this PR, but you will in the future be able to annotate a type as being `local immediate`, which will disable the mode-crossing for it.

riaqn:
@stedolan yes that's accurate summary of this patch. I have also incorporated the two last points you raised.

testsuite/tests/typing-local/local.ml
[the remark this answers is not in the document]
stedolan: How about:
```
let testbool1 f =
  let local_ r = ref 42 in
  (f r || false) && true

let testbool2 f =
  let local_ r = ref 42 in
  true && (false || f r)
```

typing/typecore.ml
```diff
@@ -2026,7 +2060,7 @@ and type_pat_aux
           pat_desc = Tpat_alias(q, id, name);
           pat_loc = loc; pat_extra=[];
           pat_type = q.pat_type;
-          pat_mode = q.pat_mode;
+          pat_mode = mode_cross !env expected_ty q.pat_mode;
```
stedolan, line 2063: I'm a little suspicious of this line. I'd expect `pat_mode` (the mode in the Typedtree pattern) to match the mode passed to `enter_variable` (the mode of the binding in the environment), like in the `Ppat_var` case. Should this call to `mode_cross` come before `enter_variable`?

typing/typecore.ml
```diff
@@ -545,6 +545,40 @@ let has_local_attr_pat ppat =
...
+        (* if !Clflags.verbose then begin
+         *   Format.eprintf "@[mode crossed!!@.%a@.@]" Printtyp.raw_type_expr  ty
+         * end; *)
+        Value_mode.newvar ()  (* floating and relaxed *)
+      end
+    | _ -> begin
+      (* if !Clflags.verbose then begin
+       *   Format.eprintf "@[mode not crossed because not immediate@.%a@.@]"
+       *     Printtyp.raw_type_expr ty
+       * end; *)
```
stedolan, line 563: Should this commented out code stay?

(I don't mind either way: leaving commented-out debug printing code can be useful if you think you'll need it soon, but if you didn't plan to keep it now is the time to delete it)

stedolan (changes requested):

typing/typecore.ml
lpw25: FWIW I think debug code should always be deleted. You can find it in an old commit if you need it again, and when it isn't deleted it tends to stay forever slowly bit-rotting. For example, see the many commented out bits of code Jacques has left in the type-checker.

goldfirere (commented):
It would be great if there were an updated to the documentation in `jane/doc` (in the repo) that came along with this patch. One question I have about the design: does this mode-crossing apply to any `immediate` type? or just `int`? Here are some examples of types that could mode-cross:

```ocaml
module M : sig
  type t [@@immediate]
end = struct
  type t = int
end

type t2 = { x : int } [@@unboxed]
```

Both `M.t` and `t2` are safe to mode-cross. But do they in practice?

Separately: is there a way to opt out of this behavior? I've already had a few folks ask me how to preserve the normal `local` behavior even in the presence of `immediate`s. They want to use `local`  to prevent escaping, not because they care about stack allocation.

It makes sense (and may indeed be preferable) to address of these in later patches, though!

typing/typecore.ml
```diff
@@ -5109,6 +5168,7 @@ and type_function ?in_function
       Final_arg { partial_mode = Alloc_mode.join [arg_mode; alloc_mode] }
     end
   in
+  let cases_expected_mode = expect_mode_cross env ty_res cases_expected_mode in
```
lpw25, line 5171: This can actually only have an effect when `uncurried_function` is `false`. We can save a little bit of work by moving this into the `else` branch above.

testsuite/tests/typing-local/crossing.ml
```diff
@@ -0,0 +1,321 @@
...
+let f : local_ foo -> _ =
+  fun r -> r.y
+[%%expect{|
+Line 2, characters 11-14:
+2 |   fun r -> r.y
+               ^^^
+Error: This value escapes its region
+|}]
+
+(* the following failed, because the expected type
```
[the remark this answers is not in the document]
lpw25: This was actually a missing case in the PR. It should work now.

lpw25:
> Both M.t and t2 are safe to mode-cross. But do they in practice?

Yes

> Separately: is there a way to opt out of this behavior? I've already had a few folks ask me how to preserve the normal local behavior even in the presence of immediates. They want to use local to prevent escaping, not because they care about stack allocation.

There is not currently a way to do it. In future you should be able to give types the kind `local immediate` to get an immediate that won't cross locality modes. I'd rather wait for the layouts branch to merge before adding this ability though.

typing/typecore.ml
lpw25, line 5171: This can actually only have an effect when `uncurried_function` is `false`. We can save a little bit of work by moving this into the `else` branch above.

stedolan:
There was a commit that removed some debug logic and made some changes to generalisation (7bbdb7c6282df3dc191f9f86b7f253db6e0a89f1), but it seems to have gotten lost in a rebase.

riaqn:
> There was a commit that removed some debug logic and made some changes to generalisation ([7bbdb7c](https://github.com/ocaml-flambda/ocaml-jst/commit/7bbdb7c6282df3dc191f9f86b7f253db6e0a89f1)), but it seems to have gotten lost in a rebase.

Yes, Leo found a better way to fix the generalisation bug, which is done in https://github.com/ocaml-flambda/ocaml-jst/pull/58/commits/f59c3a56a21e917be5bc31ba39cea6c437ef0953

stedolan:
> Yes, sorry I forgot to mention that Leo found a better way to fix the generalisation bug, which is done in https://github.com/ocaml-flambda/ocaml-jst/commit/f59c3a56a21e917be5bc31ba39cea6c437ef0953

Right, that explains why those earlier generalisation changes are no longer needed, but the debug code has reappeared because that commit was dropped.

typing/typecore.ml
```diff
@@ -3775,15 +3794,24 @@ and type_expect_
...
           match get_desc ty with
             Tarrow (_l, ty_arg, ty_fun, _com) ->
               (try unify_var env (newvar()) ty_arg
                with Unify _ -> assert false);
-              lower_args (TypeSet.add ty seen) ty_fun
-          | _ -> ()
+              ret_tvar (TypeSet.add ty seen) ty_fun
+          | Tvar _ -> 
+              let v = newvar () in 
+              let rt = get_level ty > get_level v in
```
stedolan, line 3808: Here `ret_tvar` is computed as `get_level ty > get_level v`, but in the old code it was `get_level ty_fun >= get_level ty_arg`. Is the change from `>` to `>=` intentional? (I don't really get what's going on with this change)

typing/typecore.ml
```diff
@@ -3740,18 +3757,20 @@ and type_expect_
...
-        type_expect ?in_function ~recarg env mode_local sbody
+        type_expect ?in_function ~recarg env mode sbody
           ty_expected_explained
       in
       { exp with exp_loc = loc }
   | Pexp_apply
       ({ pexp_desc = Pexp_extension({txt = ("ocaml.local" | "local")}, PStr []) },
        [Nolabel, sbody]) ->
-      submode ~loc ~env Value_mode.local expected_mode;
+      let mode = expect_mode_cross env ty_expected mode_local in
```
stedolan, line 3770: I don't understand what this case is for - when we're already at `mode_local`, what does mode crossing do?

stedolan (changes requested):

typing/typecore.ml
riaqn: My understanding is that, the original code enforces `local_ 42` to be local. With this modification, however, `local_ 42` would be unconstrained (and could be global)

stedolan: Thanks for the explanation! I now see that the change accepts more programs and is sound, but I'm still not sure I actually want it - it seems confusing that `ref (local_ 42)` is accepted. Is there a use-case I'm missing here?

riaqn: Yeah Leo and I wasn't sure either. But if we stick to the principle that "immediates don't respect mode", we should probabaly allow `ref (local_ 42)` to type check?

typing/typecore.ml
riaqn: yes it is intentional, because `get_level v` is one less now, because the original code has L3825

typing/typecore.ml
goldfirere: As something of an outside observer, I'd expect `ref (local_ 42)` to be just fine: immediates simply don't care about locality. That seems simpler than trying (somehow) to avoid this.

stedolan: OK, let's go with this as is!

typing/typecore.ml
stedolan: Thanks! Missed that.

stedolan (approved):

PR #59, opened 2022-11-01
Add/move some documentation

goldfirere:
As I was reviewing #46, I made a few improvements (hopefully) to documentation, posted here.

### DIFF
--- a/typing/typeopt.ml
+++ b/typing/typeopt.ml
@@ -21,6 +21,9 @@ open Asttypes
 open Typedtree
 open Lambda
 
+(* Expand a type, looking through ordinary synonyms, private synonyms,
+   links, and [@@unboxed] types. The returned type will be therefore be none
+   of these cases. *)
 let scrape_ty env ty =
   let ty = Ctype.expand_head_opt env (Ctype.correct_levels ty) in
   match ty.desc with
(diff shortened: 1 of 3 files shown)


typing/typeopt.ml
```diff
@@ -59,13 +63,15 @@ let maybe_pointer_type env ty =
...
-  | Int
+  | Int   (* any immediate type *)
   | Float
   | Lazy
   | Addr  (* anything except a float or a lazy *)
   | Any
 
-let classify env ty =
+(* Classify a ty into a [classification]. Looks through synonyms, using [scrape_ty].
+   Returning [Any] is safe, though may skip some optimizations. *)
```
antalsz, line 73: Looks like this comment about `Any` got lost, and is probably worth keeping around in some form?

antalsz (changes requested):
I like these changes!  Had one note about something that seems to have gotten lost, but other than that, LGTM

typing/typeopt.ml
goldfirere: That comment about `Any` is new here. Maybe you somehow diff'd backwards?

goldfirere:
Landed as https://github.com/ocaml-flambda/ocaml-jst/commit/8788ff6c839318c8876e89888c95297dd7784ee7

PR #60, opened 2022-11-02
Move `global_flag` to `Asttypes`

antalsz:
When `type global_flag` was just in the `.mly` file, Merlin had trouble, as Merlin wants to export the various tokens and nonterminals, but that type is not exportable.  It's easier to have it live in `Asttypes`, and arguably makes more sense, anyway

### DIFF
--- a/boot/menhir/parser.ml
+++ b/boot/menhir/parser.ml
@@ -16,7 +16,7 @@ module MenhirBasics = struct
     | VAL
     | UNDERSCORE
     | UIDENT of (
-# 843 "parsing/parser.mly"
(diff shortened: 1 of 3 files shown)


goldfirere:
While the content certainly looks innocent enough to me (and a modest improvement), I'm not sure about the CI story here. The only thing that might go wrong is a failure to compile. How do we know that hasn't happened? Is there any CI set up?

antalsz:
@goldfirere We lost the CI because I accidentally did this against 4.12 and not 4.14; I'm rebasing

goldfirere (approved):

PR #61, opened 2022-11-03
Propagate location information to `local_` in expressions

antalsz:
This information was updated to propagate to `local_` in types and patterns, where it's encoded as an attribute, but not in expressions, where it's encoded as an extension node.  Getting the location information in types and patterns is/was a blocker for some automated refactorings; getting the location information out further should be similarly useful.  I put this against 4.14 since we'll be rolling forward to that soon.


goldfirere (approved):
LGTM

lpw25:
This needs a rebase. Then it should be merged.

PR #62, opened 2022-11-04
more entries in gitignore

riaqn:
Don't know if it's a 4.14 thing, but some build files are not included in the current `.gitignore` file and could be annoying and slowing down vscode.


antalsz:
Nice, thanks!

PR #63, opened 2022-11-14
Add debug_printers.ml

lpw25:
Add some helper files to make it easier to install useful printers when debugging the compiler with ocamldebug.


stedolan:
This looks great, but could you do one for Path.t as well?

lpw25:
Added path

PR #64, opened 2022-11-18
Enable support for Jane Street's internal Merlin configuration

antalsz:
Now Jane Street internal developers can get Merlin within the compiler when building with `dune`.

Once we improve the internal tooling in a couple of ways (see comments) we can simplify some of the way this is done.

@stedolan: Why is `dune-project` only created from `dune-project.jst`?

### DIFF
--- a/jane-street-merlin-setup.sh
+++ b/jane-street-merlin-setup.sh
@@ -12,17 +12,16 @@ fi
 set -e
 
 merlin_dir="$(dirname "$ocamlmerlin")"
+echo "$merlin_dir" > .local-merlin-binaries
+ocamlc -where > .local-ocaml-lib
 
-# The stub files `jenga.conf` and `.ocaml-lib` should be deleted if
-# `.local-merlin-binaries` exists or if the `.merlin-binaries` search ever stops
-# looking for a tree file first.
+# When `.local-merlin-binaries` is fully supported (rather than just supported
+# in dev Emacs), we can drop the extra directory `.for-jane-street-merlin` as
+# well as the stub `jenga.conf` file.
 
-# When `.local-merlin-binaries` is fully supported we can just point it at
-# `$MERLIN_DIR` and drop this extra directory *and* the stub files entirely.
-# This will require updating the `.gitignore`, too.
 mkdir -p .for-jane-street-merlin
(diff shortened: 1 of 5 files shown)


stedolan:
> @stedolan: Why is dune-project only created from dune-project.jst?

In flambda-backend, the contents of ocaml-jst are imported as `ocaml/`, and having two dune-project files confuses dune. So ocaml-jst's `dune-project` is only created if you're building using ocaml-jst's build.

stedolan:
Thanks for this! Looks good.

Why is the merlin setup script part of the `$(dune)` variable? Does it need to be run every time dune is invoked?

jane-street-merlin-setup.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# When `.local-merlin-binaries` is fully supported we can just point it at
+# `$MERLIN_DIR` and drop this extra directory *and* the stub files entirely.
+# This will require updating the `.gitignore`, too.
+mkdir -p .for-jane-street-merlin
+ln -sfn "$merlin_dir" .for-jane-street-merlin/dev
+ln -sfn "$merlin_dir" .for-jane-street-merlin/prod
+
+echo "$PWD/.for-jane-street-merlin" > .merlin-binaries
+ocamlc -config | grep standard_library: | cut -d' ' -f2 > .ocaml-lib
```
stedolan, line 28: This string is also available as `ocamlc -where`.

antalsz:
@stedolan: I set it up to run as part of `dune` because I figured that would keep things up to date in the faces of changes.  But now I'm not 100% sure about changes to *what* -- changes to the ambient `ocamlc`, I guess?  Regardless, though, it seemed cheap to run and means you don't have to think about the structure of things to get access to Merlin.  I'm on board with moving this to a good start rule if you want to, though; feel free to move it there and then click "merge", or just click "merge" without that!

antalsz:
@stedolan Does this latest push fix your problem?

lpw25:
@stedolan ping. I think it would be good to get this situation resolved.

@antalsz Do we also need a similar patch for `flambda-backend`?

stedolan:
> @stedolan Does this latest push fix your problem?

It does indeed! Thanks!

PR #70, opened 2022-11-28
Ensure that Ctype.nongen always calls remove_mode_variables

### LINKED ISSUE #65: Local inference and class inheritance bug
The opam package `mm` fails to build with the flambda-backend compilers, with a type error.
I managed to create a small example and reproduce the issue with the `ocaml-jst` repo, so I'm creating an issue here.

The small example has two files, `dep.ml` and `main.ml` (no mli files).
`Dep` contains class definitions:
```ocaml
class c =
  object
    method private m () () = 0
  end

class virtual cv =
  object
    method virtual private m : unit -> unit -> int
  end
```
`Main` contains one class inheriting from those:
```ocaml
class c fname =
  object
    inherit Dep.c
    inherit Dep.cv
  end
```
Then compiling `main.ml` (after compiling `dep.ml`) yields the following error:
```
4 |     inherit Dep.cv
        ^^^^^^^^^^^^^^
Error: The method m has type unit -> unit -> int but is expected to have type
         unit -> local_ (unit -> int)
       Type unit -> unit -> int is not compatible with type
         unit -> local_ (unit -> int) 
```

stedolan:
`Ctype.nongen` checks that there are no weakly polymorphic variables, before e.g. inferring the module type of a compilation unit that does not have an mli file, or using `module type of`.

With locals, calls to this function must also be accompanied by a call to `Ctype.remove_mode_variables`, which ensures that there are no unresolved mode variables.

One of these calls got lost in the 4.14 merge, causing #65. The fix here is to make `nongen` call `remove_mode_variables` directly so that this can't happen.

(Fixes #65)

### DIFF
--- a/testsuite/tests/typing-local/regression_class_dep.ml
+++ b/testsuite/tests/typing-local/regression_class_dep.ml
@@ -1,0 +1,9 @@
+class c =
+  object
+    method private m () () = 0
+  end
+
+class virtual cv =
+  object
+    method virtual private m : unit -> unit -> int
+  end
--- a/testsuite/tests/typing-local/regression_class_type.ml
+++ b/testsuite/tests/typing-local/regression_class_type.ml
@@ -1,0 +1,18 @@
+(* TEST
+readonly_files = "regression_class_dep.ml"
+* setup-ocamlc.opt-build-env
+** ocamlc.opt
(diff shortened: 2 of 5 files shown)


lpw25 (approved):
LGTM

PR #71, opened 2022-11-30
Test unboxed numerics (part of unboxed-types version 2)

goldfirere:
@lpw25 and I thought it might be good to write tests to accurately define the expected functionality of upcoming releases of unboxed types. There doesn't seem to be a good reason not to just check this into the repo now, but I'm happy to consider other ways of doing this.


goldfirere:
Chatted about this with @ceastlund, who may enjoy taking a look.

testsuite/tests/typing-layouts/unboxed-ints.ml
```diff
@@ -0,0 +1,270 @@
...
+
+let x = Uint64.(#3L + #4L)
+let x = Uint32.(#3l + #4l)
+let x = Unativeint.(#3n + #4n)
+
+[%%expect {|
+success
+|}]
+
+let bad = #3L + #4L  (* wrong + *)
```
antalsz, line 230:
```suggestion
(* wrong + *)
let bad = #3L + #4L
```

antalsz (commented):
This basically looks good to me!  A few comments made inline

goldfirere:
These tests appear to be well represented in the current flambda-backend. Closing

PR #72, opened 2022-11-30
Run "misplaced attributes" check when compiling mlis

ccasin:
The new system that checks for unused attributes wasn't running on mli files - only ml files.  This fixes that.

A couple other changes were needed:
- There were some unused `[@inline]` attributes in the stdlib that I deleted.
- The check wasn't correctly handling `[@@@deprecated]`, and I hadn't noticed because this is only used in mlis.  Attributes like this, which are not "used" when compiling the modules they appear in but rather by clients of those modules, require special handling.


goldfirere (approved):
LGTL

PR #74, opened 2022-12-02
Better error message for under-applied functions

goldfirere:
Fixes Jira ticket TANDC-1742.

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -94,6 +94,10 @@ type existential_restriction =
   | In_class_def  (** or in [class c = let ... in ...] *)
   | In_self_pattern (** or in self pattern *)
 
+type submode_reason =
+  | Application of type_expr
+  | Other
+
 type error =
   | Constructor_arity_mismatch of Longident.t * int * int
   | Label_mismatch of Longident.t * Errortrace.unification_error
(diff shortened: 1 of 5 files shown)


typing/typecore.ml
```diff
@@ -6824,6 +6832,44 @@ let escaping_hint reason (context : Env.escaping_context option) =
...
+          if n = 0
+          then None
+          else begin match Alloc_mode.check_const alloc with
+            | Some Local -> None
+            | Some Global -> assert false (* this should have been caught above *)
+            | None -> Some (n, false)
+          end
+      in
+      loop (Amode Local) (* doesn't matter *)
+           0 ty
```
stedolan, line 6860: I found the `alloc` parameter logic in `loop` a little fiddly, so here's a suggestion for an alternative:
```ocaml
let rec loop sureness n ty =
  match get_desc ty with
  | Tarrow ((_, _, res_mode), _, res_ty, _) ->
    begin match Alloc_mode.check_const res_mode with
    | Some Global ->
      Some (n+1, true)
    | (None | Some Local) as res_mode ->
      let sureness = sureness && Option.is_some res_mode in
      loop sureness (n+1) res_ty
    end
  | _ ->
    if n = 0
    then None
    else Some (n, sureness)
in
loop true 0 ty
```

goldfirere: Perfect. Thanks!

goldfirere:
OK. I've taken @stedolan's suggestions and rebased. Anything stopping merge?

PR #75, opened 2022-12-07
Fix interaction of or-patterns, GADTs and as-patterns (#11799)

lpw25:
Backport of [#11799](https://github.com/ocaml/ocaml/pull/11799) upstream


goldfirere:
I confirm that the payload here is the same as the payload in [#11799](https://github.com/ocaml/ocaml/pull/11799). There is some debate there on whether this is the best approach, but my read of the room there is that the upstream patch will be accepted. I think you can merge here but should follow upstream if things go a different direction there.

mshinwell:

PR #76, opened 2022-12-07
Enable individual testing with Makefile.jst

goldfirere:


antalsz:
This looks good, but why do we need this over and above `make -C testsuite`?  And `make -C testsuite` provides a few more targets; it would be good to have those too.  (We don't really use `LIST=`, but I use `test-$PREFIX` and even more so `parallel-$PREFIX` all the time.)  That said, I'd approve this as-is, it looks good to me.

antalsz (approved):

goldfirere:
This works with the new dune-based build system. Without this change, you would have to build using the old build system in order to run any less than the entire testsuite. In conversation, @stedolan and I agreed that there were even further improvements beyond just this possible, but this was what I needed to move forward.

goldfirere:
This seems to work and there appears to be little downside to merging. Still happy for review from @stedolan.

stedolan:
Thanks! Looks good.

PR #78, opened 2022-12-13
Import flambda-backend patches 2022-12-13

mshinwell:
This is what I got when I ran the `pull_backend_patches` script, which I did in preparation for a merge the other way (from `ocaml-jst` to `flambda-backend`) prior to tagging 4.14.0-0.

I had to re-do the bootstrap.

There was also a build failure in `otherlibs/dynlink/` which we've also seen on the `flambda-backend` 32-bit builds.  This happens because the `dynlink_compilerlibs.cmi` inferred pack interface is written by both `ocamlc` and `ocamlopt` with different hashes, almost certainly due to sharing differences.  I've added a hack which will hopefully work robustly, together with an explanatory comment.  I also deleted a stale confusing comment nearby.


otherlibs/dynlink/Makefile
```diff
@@ -193,20 +193,22 @@ $(LOCAL_SRC)/%.cmx: $(LOCAL_SRC)/%.ml
 
 # Rules for building the [Dynlink_compilerlibs] bytecode and native packs
 # from their components.
-
+# The dynlink_compilerlibs.cmi file is unfortunately written by both the ocamlc
+# and ocamlopt invocations.  The resulting files are different, probably because
+# of sharing differences.  Writing an .mli file would be problematic.  Instead
+# we allow the compiler to generate the inferred .cmi upon the ocamlc
+# invocation, then trick it (by touching dynlink_compilerlibs.cmi) upon the
```
gretay-js, line 200:
```suggestion
# invocation, then trick it (by touching dynlink_compilerlibs.mli) upon the
```

gretay-js (approved):
I only  looked at the last "manual" commits and they are fine.

PR #82, opened 2022-12-15
Unboxed types version 3 tests

goldfirere:
This covers the addition of `or_null`, which comprises v3 of the unboxed-types plan. This test is skipped (and so can be merged now), but someone should review this, as it is intended to include the entire specification for the `or_null` feature.

### DIFF
--- a/testsuite/tests/typing-layouts/or_null.ml
+++ b/testsuite/tests/typing-layouts/or_null.ml
@@ -38,25 +38,28 @@ end
 
 module Or_null : Or_null = Or_null
 
+(* CR layouts (v3): check output to see how bad the pretty-printing is.
+   In particular, it would be nice to suppress layout annotations that
+   are implied by the rest of the signature, but this may be hard. *)
 [%%expect {|
 success
 |}]
 
 (* ensure that immediacy "looks through" or_null *)
-type t = (int or_null : immediate)
-type t = (bool or_null : immediate)
+type t1 : immediate = int or_null
+type t2 : immediate = bool or_null
 
 [%%expect {|
 success
 |}]
 
-type t = (string or_null : immediate)
+type t : immediate = string or_null
 
 [%%expect {|
 error
 |}]
 
-type t = (string or_null : value)
+type t : value = string or_null
 
 [%%expect {|
 success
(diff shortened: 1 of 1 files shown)


lpw25:
Here's one we could consider:
```ocaml
let foo x = (x : int list :> int or_null list)
```

goldfirere:
> Here's one we could consider:
> 
> ```ocaml
> let foo x = (x : int list :> int or_null list)
> ```

Great idea. I've added this.

testsuite/tests/typing-layouts/or_null.ml
```diff
@@ -0,0 +1,326 @@
...
+type t = (int or_null : value)
+
+[%%expect {|
+success
+|}]
+
+(* magic looking-through of [or_null] can't be abstracted over *)
+type 'a t = 'a or_null
+type q = (string t : value)
+type q = (int t : immediate)  (* but t isn't abstract, so this is OK *)
```
ccasin, line 100: annotation syntax

testsuite/tests/typing-layouts/or_null.ml
```diff
@@ -0,0 +1,326 @@
...
+
+module M : sig
+  type 'a t
+end = struct
+  type 'a t = 'a or_null
+end
+
+[%%expect {|
+error
+|}]
```
ccasin, line 158: Maybe a note here that the error message had better be very good (perhaps with an explicit hint) because needing a `value` annotation is unusual.

testsuite/tests/typing-layouts/or_null.ml
```diff
@@ -0,0 +1,326 @@
...
+
+type 'a q = 'a t
+type ('a : immediate) q = 'a t
+
+[%%expect {|
+success
+|}]
```
ccasin, line 123: Probably note that `q` should be immediate, or add an annotation.

ccasin (changes requested):
Looks good - a few minor comments, and I think several spots need to be updated for the current thinking on layout annotations.

testsuite/tests/typing-layouts/or_null.ml
```diff
@@ -0,0 +1,326 @@
...
+  type 'a t = 'a or_null =
+    | None
+    | Some of 'a
+
+  val none : 'a or_null
+  val some : 'a -> 'a or_null
+  val value : 'a or_null -> default:'a -> 'a
+  val get : 'a or_null -> 'a
+  val bind : 'a or_null -> ('a -> 'b or_null) -> 'b or_null
+  (* unlike [option] we cannot have [join] *)
```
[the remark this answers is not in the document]
goldfirere: I lean against, for now at least.

ccasin:
Thanks.  Merging despite the failing i386 test because that can't possibly be related.

PR #84, opened 2022-12-20
Remove a tiny code stutter I came across

goldfirere:
I saw some code, and it looked to me as if it did nothing at all. So I removed it. And it turns out it did nothing. Here's the patch.


antalsz:
This is clearly right, merged!  I wonder if it's worth bringing this upstream?

PR #85, opened 2022-12-22
Document the debug_printers script

goldfirere:
Thanks to Chris for explaining this to me!

### DIFF
--- a/HACKING.jst.adoc
+++ b/HACKING.jst.adoc
@@ -54,3 +54,19 @@ where the test file or test dir are specified with respect to the
     $ make -f Makefile.jst test-one TEST=typing-local/local.ml
     $ make -f Makefile.jst test-one DIR=typing-local
 
+## Debugging
+
+OCaml 4.14 makes `type_expr` abstract, and thus normal debug printing
+of types no longer works. However, there is now an installable printer
+for types, which we can use to see the types. Here are the instructions:
+
+1. Use the old `Makefile`, not the new `Makefile.jst`. This is an infelicity
+we hope to fix.
+
+2. In the `tools` directory, run `make debug_printers.cmo`.
+
+3. In the debugger, execute some instructions, with e.g. `run` or `step`. This forces
+the debugger to load the compiler code, required for the next
+step.
+
+4. Execute `source tools/debug_printers` to install the printers.


HACKING.jst.adoc
```diff
@@ -54,3 +54,18 @@ where the test file or test dir are specified with respect to the
...
+of types no longer works. However, there is now an installable printer
+for types, which we can use to see the types. Here are the instructions:
+
+1. Use the old `Makefile`, not the new `Makefile.jst`. This is an infelicity
+we hope to fix.
+
+2. In the `tools` directory, run `make debug_printers.cmo`.
+
+3. In the debugger, get to your breakpoint. (You must actually have
+executed some code for the next step to work.)
```
antalsz, line 69: I think it's worth being more precise here.  Does just `start` or the equivalent work?

antalsz (changes requested):
This looks great, thanks Richard!  I had one question about the exact requirements but otherwise looks good.

HACKING.jst.adoc
goldfirere: I don't think `start` is what you want (it runs the current function backwards). I use `run`.

antalsz: Oh right, it's "go to the `start`".  Yeah, that's not what I meant.  But I meant, you don't need to reach a breakpoint, right?  Just in the code somewhere?  So wouldn't `step` work too, or something else smaller?  `run` does seem like a good default, though

antalsz: I think I'd be happy with, in the parenthetical something like "(Anything else that will have loaded all the compiler code will work; this is required for the next step.)"  For instance, your instructions talk about a breakpoint but don't set one, and it's perfectly fine to run `source tools/debug_printers` when `run` has finished before running backwards.

ccasin: Taking one step seems to be enough, FWIW.  The main thing is that the search path seems to get initialized.  E.g., when ocaml debug has just started:

```
(ocd) directory
Reinitialize directory list ? (y or n) n
Directories:  . /home/ccasinghino/ocaml/layouts/ocaml-jst/out/lib/ocaml
```

and after taking one step:

```
(ocd) directory
Reinitialize directory list ? (y or n) n
Directories:  . /home/ccasinghino/ocaml/layouts/ocaml-jst/out/lib/ocaml
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/boot
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/bytecomp
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/compilerlibs
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/driver
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/file_formats
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/lambda
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/parsing
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/stdlib
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/stdlib/stdlib
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/typing
  /usr/local/home/ccasinghino/ocaml/layouts/ocaml-jst/utils
  /usr/local/home/ccasinghino/ocaml/layouts/runtime/caml
```

antalsz:
Delighted to merge this now that we all understand the debugger :-)

PR #92, opened 2022-12-28
Bootstrap

mshinwell:
This should fix the inconsistent assumptions errors with `make` builds using upstream-style build commands.


goldfirere (approved):
Presuming you're not actively trying to introduce a [compiler exploit](https://wiki.c2.com/?TheKenThompsonHack), I'm sure this is a good fix.

PR #96, opened 2023-01-03
Refactor the treatment of type variables

goldfirere:
It seemed to me that the code for handling type variable environments was distributed across a bunch of space within Typetexp. So I gathered it all together and modularized it. There should be no change in behavior.

This PR comprises 5 commits, each expressing an independent idea. We can pick and choose which one(s) we want to keep.

If this looks good locally, I will upstream.

### DIFF
--- a/typing/typecore.ml
+++ b/typing/typecore.ml
@@ -3147,7 +3147,7 @@ let rec approx_type env sty =
           (* Polymorphic types will only unify with types that match all of their
            polymorphic parts, so we need to fully translate the type here
            unlike in the monomorphic case *)
-          Typetexp.transl_simple_type env false arg_mode arg_sty
+          Typetexp.transl_simple_type env ~fixed:false arg_mode arg_sty
         in
         let ret = approx_type env sty in
         let marg = Alloc_mode.of_const arg_mode in
@@ -3182,7 +3182,7 @@ let type_pattern_approx env spat ty_expected =
         else Alloc_mode.Global
       in
       let ty_pat =
-        Typetexp.transl_simple_type env false arg_type_mode sty
+        Typetexp.transl_simple_type env ~fixed:false arg_type_mode sty
       in
       begin try unify env ty_pat.ctyp_type ty_expected with Unify trace ->
         raise(Error(spat.ppat_loc, env, Pattern_type_clash(trace, None)))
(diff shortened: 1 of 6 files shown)


typing/typetexp.ml
```diff
@@ -176,6 +159,42 @@ end = struct
   let remember_used name v loc =
     used_variables := TyVarMap.add name (v, loc) !used_variables
 
+
+  type flavor = Unification | Universal
+  type extension = Extensible | Fixed
```
antalsz, line 193: `extensibility`?

typing/typetexp.mli
```diff
@@ -17,23 +17,32 @@
...
+
+  type poly_univars
+  val make_poly_univars : string list -> poly_univars
+    (* Create a set of univars with given names *)
+  val check_poly_univars :
+     Env.t -> Location.t -> poly_univars -> type_expr list
+    (* Verify that the given univars are universally quantified,
+       and return the list of variables. The type in which the
+       univars are used must be generalised *)
+end
```
antalsz, line 37: I don't really understand how you chose which things to put in this module

antalsz (approved):
I think this looks great!  I had two notes:

1. I think `narrow_in` is an awkward name; I'd prefer `narrow`, `narrowed`, or `narrowing`.  If you still prefer `narrow_in`, that's fine; no need to go back and forth, I'm happy.
2. I don't really understand what's in `TyVarEnv` vs. what isn't.

But I think this is a good set of improvements and all 5 commits could go upstream :+1:

goldfirere:
This is a test message to see whether I get a notification.

typing/typecore.ml
```diff
@@ -6076,33 +6077,33 @@ and type_unpacks ?(in_function : (Location.t * type_expr * bool) option)
...
-          md_loc = unpack.tu_name.loc;
-          md_uid = unpack.tu_uid; }
-      in
-      let (id, env) =
-        Env.enter_module_declaration ~scope ~shape:md_shape
-          unpack.tu_name.txt pres md env
-      in
-      Typetexp.widen ();
-      env, (id, unpack.tu_name, pres, modl) :: tunpacks
+      Typetexp.narrow_in begin fun () ->
```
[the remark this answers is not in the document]
goldfirere: I like the explicitness of marking off a region of code that is in one of these scoping constructs. Parentheses are used within expressions and seem to have a more local extent. It's all just style. I've also seen e.g. `narrow_in @@ fun () ->`, but I just find `@@` ugly. I don't see much of a consistent style around this (or other things, really) in the code, so I just went with what I liked most. Push back if you like. :)

typing/typetexp.mli
goldfirere: These happen to be the functions used externally. No other logic here. I've added a small comment to answer your reasonable question.

PR #97, opened 2023-01-09
Make immediate64 things mode cross

ccasin:
This is only safe if we're sure we don't run bytecode on 64-bit platforms.

### DIFF
--- a/testsuite/tests/typing-local/crossing.ml
+++ b/testsuite/tests/typing-local/crossing.ml
@@ -343,4 +343,16 @@ module M : sig type t [@@immediate] end
 type t2 = { x : int; } [@@unboxed]
 val f : local_ M.t -> M.t = <fun>
 val f : local_ t2 -> t2 = <fun>
-|}]  
+|}]
+
+
+(* Mode crossing works on immediate64 types *)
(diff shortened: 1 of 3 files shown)


lpw25:
I think you mean 32-bit platforms. The issue actually only happens if you build on 64-bit platforms and then run on 32-bit platforms, which we definitely don't do. We'll use a more principled fix later once layouts make that easier to do.

lpw25 (approved):

ccasin:
> I think you mean 32-bit platforms. The issue actually only happens if you build on 64-bit platforms and then run on 32-bit platforms, which we definitely don't do. We'll use a more principled fix later once layouts make that easier to do.

Yeah, typo, sorry.

lpw25:
I've pushed a change to run the test only on 64bit. That should hopefully fix the CI.

stedolan:
> I think you mean 32-bit platforms. The issue actually only happens if you build on 64-bit platforms and then run on 32-bit platforms, which we definitely don't do. We'll use a more principled fix later once layouts make that easier to do.

More specifically, things are bad only if you build on 64-bit, and run on 32-bit *with stack allocation enabled*. We could just decide that stack allocation is only supported on 64 bit configurations?

PR #98, opened 2023-01-13
Catch misplaced attributes inserted by ppx (and a little attribute cleanup)

ccasin:
Two commits:
- The first fixes a bug in the new misplaced attributes system - if a compiler built-in attribute was inserted by a ppx, we wouldn't check whether it was misplaced.  Now we do, and I added a corresponding test case with a tiny ppx.  I track whether warnings are inserted by rewriter so that the warning can mention this, since otherwise it would just mention an attribute that doesn't appear in the source.
- The second cleans up the treatment of the "noalloc" and "poll" attributes in `Translattribute` to use the new system.  I also added misplaced attributes tests for `poll`.

### DIFF
--- a/parsing/ast_invariants.ml
+++ b/parsing/ast_invariants.ml
@@ -170,6 +170,10 @@ let iterator =
           "In object types, attaching attributes to inherited \
            subtypes is not allowed."
   in
+  let attribute self attr =
+    super.attribute self attr;
+    Builtin_attributes.register_attr ~from_ppx:true attr.attr_name
+  in
   { super with
     type_declaration
   ; typ
@@ -185,6 +189,7 @@ let iterator =
   ; signature_item
   ; row_field
   ; object_field
+  ; attribute
   }
 
 let structure st = iterator.structure iterator st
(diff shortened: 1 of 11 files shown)


gretay-js (commented):
Thanks for fixing the treatment on `noalloc` attribute, that part looks fine to me.

parsing/ast_invariants.ml
```diff
@@ -170,6 +170,10 @@ let iterator =
           "In object types, attaching attributes to inherited \
            subtypes is not allowed."
   in
+  let attribute self attr =
+    super.attribute self attr;
+    Builtin_attributes.register_attr ~from_ppx:true attr.attr_name
```
goldfirere, line 175: This `~from_ppx:true` surprises me. Why is it always from a ppx here?

ccasin: Nice catch.  My initial read of the code was that these invariant checkers are only run on code produced by a ppx.  Looking closer at `pparse.ml`, I see I missed one other case: they are also run if we're parsing an "ast_file" (I think this is just the output of `-dparsetree`).  An obvious solution is to just add a parameter indicating whether we're running these on the output of a ppx or not, but I'd like to check to see if we're ever using ast files for anything first (if so, perhaps there should be three versions of the error instead of two).

PR #99, opened 2023-01-18
Refactor layouts not to use `repr`

goldfirere:
This commit refactors the layouts mechanism for looking through sort variables to be more like 4.14, where it's structurally impossible to forget to do so.

For better or worse (worse, probably), it's a diff on my layout-annotations patch. But worth reviewing separately, and it doesn't seem worth the time to fix this infelicity.

@ccasin to review. All the action is in types.ml(i), with a bunch of code moved from type_layout.ml(i). The `intersection` and `sublayout` functions remain in type_layout and deserve a careful check, as they've been rewritten.

Extra bonus: this patch puts `immediate <: immediate64`, not the other way around.


goldfirere:
Subsumed by https://github.com/ccasin/ocaml-jst/pull/2

PR #100, opened 2023-01-18
remove exp_mode from typedtree

riaqn:
This patch tries to remove `exp_mode` from `typedtree`, which is hotbed for subtle mode bugs. Instead, we only keep `alloc_mode` in some typed expressions where allocations happen.

### DIFF
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -98,10 +98,6 @@ let transl_alloc_mode alloc_mode =
   | Global -> alloc_heap
   | Local -> alloc_local
 
-let transl_exp_mode e =
-  let alloc_mode = Value_mode.regional_to_global_alloc e.exp_mode in
-  transl_alloc_mode alloc_mode
-
 let transl_apply_position position =
   match position with
   | Default -> Rc_normal
@@ -194,12 +190,12 @@ let rec push_defaults loc bindings use_lhs cases partial warnings =
   match cases with
     [{c_lhs=pat; c_guard=None;
       c_rhs={exp_desc = Texp_function { arg_label; param; cases; partial;
-                                        region; curry; warnings } }
+                                        region; curry; warnings; alloc_mode } }
         as exp}] when bindings = [] || trivial_pat pat ->
       let cases = push_defaults exp.exp_loc bindings false cases partial warnings in
       [{c_lhs=pat; c_guard=None;
         c_rhs={exp with exp_desc = Texp_function { arg_label; param; cases;
-          partial; region; curry; warnings }}}]
+          partial; region; curry; warnings; alloc_mode }}}]
   | [{c_lhs=pat; c_guard=None;
       c_rhs={exp_attributes=[{Parsetree.attr_name = {txt="#default"};_}];
              exp_desc = Texp_let
@@ -362,15 +358,15 @@ and transl_exp0 ~in_new_scope ~scopes e =
       transl_let ~scopes rec_flag pat_expr_list
         body_kind (event_before ~scopes body (transl_exp ~scopes body))
   | Texp_function { arg_label = _; param; cases; partial;
-                    region; curry; warnings } ->
+                    region; curry; warnings; alloc_mode } ->
       let scopes =
         if in_new_scope then scopes
         else enter_anonymous_function ~scopes
       in
-      transl_function ~scopes e param cases partial warnings region curry
+      transl_function ~scopes e alloc_mode param cases partial warnings region curry
   | Texp_apply({ exp_desc = Texp_ident(path, _, {val_kind = Val_prim p},
                                        Id_prim pmode);
-                exp_type = prim_type } as funct, oargs, pos)
+                exp_type = prim_type; } as funct, oargs, pos, alloc_mode)
     when can_apply_primitive p pmode pos oargs ->
       let argl, extra_args = cut p.prim_arity oargs in
       let arg_exps =
@@ -393,20 +389,18 @@ and transl_exp0 ~in_new_scope ~scopes e =
         let tailcall = Translattribute.get_tailcall_attribute funct in
         let inlined = Translattribute.get_inlined_attribute funct in
         let specialised = Translattribute.get_specialised_attribute funct in
-        let e = { e with exp_desc = Texp_apply(funct, oargs, pos) } in
         let position = transl_apply_position pos in
-        let mode = transl_exp_mode e in
+        let mode = transl_alloc_mode alloc_mode in
         event_after ~scopes e
           (transl_apply ~scopes ~tailcall ~inlined ~specialised ~position ~mode
              lam extra_args (of_location ~scopes e.exp_loc))
       end
-  | Texp_apply(funct, oargs, position) ->
+  | Texp_apply(funct, oargs, position, alloc_mode) ->
       let tailcall = Translattribute.get_tailcall_attribute funct in
       let inlined = Translattribute.get_inlined_attribute funct in
       let specialised = Translattribute.get_specialised_attribute funct in
-      let e = { e with exp_desc = Texp_apply(funct, oargs, position) } in
       let position = transl_apply_position position in
-      let mode = transl_exp_mode e in
+      let mode = transl_alloc_mode alloc_mode in
       event_after ~scopes e
         (transl_apply ~scopes ~tailcall ~inlined ~specialised
            ~position ~mode (transl_exp ~scopes funct)
@@ -420,17 +414,17 @@ and transl_exp0 ~in_new_scope ~scopes e =
                Matching.for_trywith ~scopes k e.exp_loc (Lvar id)
                  (transl_cases_try ~scopes pat_expr_list),
                Typeopt.value_kind e.exp_env e.exp_type)
-  | Texp_tuple el ->
+  | Texp_tuple (el, alloc_mode) ->
       let ll, shape = transl_list_with_shape ~scopes el in
       begin try
         Lconst(Const_block(0, List.map extract_constant ll))
       with Not_constant ->
         Lprim(Pmakeblock(0, Immutable, Some shape,
-                         transl_exp_mode e),
+                         transl_alloc_mode alloc_mode),
               ll,
               (of_location ~scopes e.exp_loc))
       end
-  | Texp_construct(_, cstr, args) ->
+  | Texp_construct(_, cstr, args, alloc_mode) ->
       let ll, shape = transl_list_with_shape ~scopes args in
       if cstr.cstr_inlined <> None then begin match ll with
         | [x] -> x
@@ -445,7 +439,7 @@ and transl_exp0 ~in_new_scope ~scopes e =
             Lconst(Const_block(n, List.map extract_constant ll))
           with Not_constant ->
             Lprim(Pmakeblock(n, Immutable, Some shape,
-                             transl_exp_mode e),
+                             transl_alloc_mode alloc_mode),
                   ll,
                   of_location ~scopes e.exp_loc)
           end
@@ -455,7 +449,7 @@ and transl_exp0 ~in_new_scope ~scopes e =
           if is_const then lam
           else
             Lprim(Pmakeblock(0, Immutable, Some (Pgenval :: shape),
-                             transl_exp_mode e),
+                             transl_alloc_mode alloc_mode),
                   lam :: ll, of_location ~scopes e.exp_loc)
       end
   | Texp_extension_constructor (_, path) ->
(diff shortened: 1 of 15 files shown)


typing/typemod.ml
```diff
@@ -2551,12 +2551,6 @@ and type_structure ?(toplevel = None) funct_body anchor env sstr =
         let () = if rec_flag = Recursive then
           Typecore.check_recursive_bindings env defs
         in
-        if Option.is_some toplevel then begin
-          (* Values bound by '_' still escape in the toplevel, because
-              they may be printed even though they are not named *)
-          defs |> List.iter (fun vb ->
-            Typecore.escape ~loc:vb.vb_pat.pat_loc ~env:newenv vb.vb_expr.exp_mode);
-        end;
```
riaqn, line 2559: this hack is removed for simplicity; causing one test in `local.ml` failing; other workaround in https://github.com/ocaml-flambda/ocaml-jst/pull/2

lambda/translcore.ml
```diff
@@ -1249,11 +1242,6 @@ and transl_let ~scopes ?(add_regions=false) ?(in_structure=false)
           Translattribute.add_function_attributes lam vb_loc vb_attributes
         in
         let lam = if add_regions then maybe_region lam else lam in
-        begin match transl_exp_mode expr, lam with
-        | Alloc_heap, _ -> ()
-        | Alloc_local, Lfunction _ -> ()
-        | _ -> Misc.fatal_error "transl_let: local recursive non-function"
-        end;
```
riaqn, line 1256: this check removed for simplicity; @stedolan suggested to restore it in backend

lambda/translcore.ml
```diff
@@ -731,7 +725,6 @@ and transl_exp0 ~in_new_scope ~scopes e =
       (* when e needs no computation (constants, identifiers, ...), we
          optimize the translation just as Lazy.lazy_from_val would
          do *)
-      assert (is_heap_mode (transl_exp_mode e));
```
riaqn, line 734: this check removed for simplicity; need to restore somewhere.

lambda/translcore.ml
riaqn: this is restored in https://github.com/ocaml-flambda/flambda-backend/pull/1081

typing/typecore.ml
```diff
@@ -3997,45 +3993,46 @@ and type_expect_
...
           | _ -> raise Exit
           end
       | _ -> raise Exit
       with Exit ->
-        let arg = Option.map (type_exp env argument_mode) sarg in
-        let arg_type = Option.map (fun arg -> arg.exp_type) arg in
+        let arg = match sarg with
+        | None -> None
+        | Some sarg -> Some (type_exp env argument_mode sarg,
+                        register_allocation expected_mode)
```
goldfirere, line 4033: This looks like it might call `register_allocation` twice, whereas the previous design only ever called it once. Are you sure this change is safe?

typing/typemod.ml
goldfirere: I don't understand. Why is this OK to remove?

goldfirere (commented):
Looks pretty good, but I have a few questions in my comments.

typing/typemod.ml
riaqn: Sorry for the confusion - it's not OK to remove, and I will ask @stedolan how to restore it somewhere else.

typing/typedtree.mli
```diff
@@ -197,7 +196,7 @@ and expression_desc =
   | Texp_function of { arg_label : arg_label; param : Ident.t;
       cases : value case list; partial : partial;
       region : bool; curry : fun_curry_state;
-      warnings : Warnings.state; }
+      warnings : Warnings.state; alloc_mode : Types.alloc_mode}
```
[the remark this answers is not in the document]
riaqn: Thanks for the reminder - comments added - it's actually the mode of the whole function.

typing/typedtree.mli
```diff
@@ -211,7 +210,7 @@ and expression_desc =
             partial_mode is the mode of the resulting closure if this function
             is partially applied to a single argument.
          *)
-  | Texp_apply of expression * (arg_label * apply_arg) list * apply_position
+  | Texp_apply of expression * (arg_label * apply_arg) list * apply_position * Types.alloc_mode
```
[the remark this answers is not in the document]
riaqn: Commented - thanks. Most `alloc_mode` describes the mode of the results, except in `Texp_setfield` it describes the mode of the record. With that exception I think it's clearer to simply comment for each case separately.

typing/typedtree.mli
```diff
@@ -263,10 +263,10 @@ and expression_desc =
               { fields = [| l1, Kept t1; l2 Override P2 |]; representation;
                 extended_expression = Some E0 }
         *)
-  | Texp_field of expression * Longident.t loc * Types.label_description
+  | Texp_field of expression * Longident.t loc * Types.label_description * Types.alloc_mode
```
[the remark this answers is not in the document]
riaqn: yep - commented.

lambda/translcore.ml
```diff
@@ -393,20 +389,19 @@ and transl_exp0 ~in_new_scope ~scopes e =
         let tailcall = Translattribute.get_tailcall_attribute funct in
         let inlined = Translattribute.get_inlined_attribute funct in
         let specialised = Translattribute.get_specialised_attribute funct in
-        let e = { e with exp_desc = Texp_apply(funct, oargs, pos) } in
+        let e = { e with exp_desc = Texp_apply(funct, oargs, pos, alloc_mode) } in
```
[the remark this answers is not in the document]
riaqn: This line similar to the following one looks redundant. I removed it in the latest commit - but maybe I should ask someone else.

typing/typecore.ml
riaqn: nice catch! Fixed.

lambda/translcore.ml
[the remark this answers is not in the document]
goldfirere: No, I'm pretty sure it's redundant. If removing it works, let's go with that. What caught me was that in one place, you removed a line like this, and in this place, you kept it -- just wanted to understand that better.

typing/typedtree.mli
```diff
@@ -238,19 +239,25 @@ and expression_desc =
...
         (** C                []
             C E              [E]
             C (E1, ..., En)  [E1;...;En]
+
+            alloc_mode is the mode of the construct (heap/stack)
          *)
-  | Texp_variant of label * expression option
+  | Texp_variant of label * (expression * Types.alloc_mode) option
+        (* alloc_mode is the mode of the variant;
+           apparantly only need that when it has argument, for otherwise
```
goldfirere, line 254:
```suggestion
           needed only when the variant has an argument, for otherwise
```

typing/typedtree.mli
```diff
@@ -262,11 +269,15 @@ and expression_desc =
             Texp_record
               { fields = [| l1, Kept t1; l2 Override P2 |]; representation;
                 extended_expression = Some E0 }
+            alloc_mode is the mode of the record (stack/heap)
         *)
-  | Texp_field of expression * Longident.t loc * Types.label_description
+  | Texp_field of expression * Longident.t loc * Types.label_description * Types.alloc_mode
+    (* alloc_mode is the mode of the result; only used when the field is float *)
```
goldfirere, line 275: Not quite: it's used only when the field is an element of a float-record. Here's a possible comment:

```
(* alloc_mode is the mode of the result; used when getting a field from a record of all floats to decide where to allocate the returned float *)
```

If a record is all floats, OCaml inlines the floats. This means that when extracting a field from such a record, we need to allocate a block for the returned float.

goldfirere (approved):

typing/typedtree.mli
riaqn: thanks - I didn't know that.

lambda/translcore.ml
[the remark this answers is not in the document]
riaqn: sounds good - removed.

typing/typemod.ml
riaqn: This issue is fixed by https://github.com/ocaml-flambda/ocaml-jst/pull/104

riaqn:
This is now rebased on https://github.com/ocaml-flambda/ocaml-jst/pull/104 and no longer have regression.

goldfirere (approved):

typing/typecore.ml
```diff
@@ -2878,7 +2880,7 @@ let type_omitted_parameters expected_mode env ty_ret mode_ret args =
              let closed_args = new_closed_args @ closed_args in
              let open_args = [] in
              let mode_closure = Alloc_mode.join (mode_fun :: closed_args) in
-             register_allocation_mode mode_closure;
+             ignore (register_allocation_mode mode_closure);
```
lpw25, line 2883: I think this `ignore` isn't needed.

typing/typecore.ml
```diff
@@ -3684,11 +3686,10 @@ and type_expect_
       ruem ~mode ~expected_mode {
         exp_desc; exp_loc = loc; exp_extra = [];
         exp_type = desc.val_type;
-        exp_mode = expected_mode.mode;
         exp_attributes = sexp.pexp_attributes;
         exp_env = env }
   | Pexp_constant(Pconst_string (str, _, _) as cst) ->
-    register_allocation expected_mode;
+    ignore (register_allocation expected_mode);
```
lpw25, line 3692: This `ignore` suggests that this call to `register_allocation` probably isn't needed.

typing/typecore.ml
```diff
@@ -4024,18 +4017,22 @@ and type_expect_
...
                    exp_env = env }
           | _ -> raise Exit
           end
       | _ -> raise Exit
       with Exit ->
-        let arg = Option.map (type_exp env argument_mode) sarg in
-        let arg_type = Option.map (fun arg -> arg.exp_type) arg in
+        let arg = match sarg with
+        | None -> None
+        | Some sarg -> Some (type_exp env argument_mode sarg,
```
lpw25, line 4032: (Style note: more usual to put a newline after `->` if the case doesn't fit on a single line.)

typing/typecore.ml
```diff
@@ -4197,11 +4192,11 @@ and type_expect_
       re {
         exp_desc = Texp_record {
             fields; representation;
-            extended_expression = opt_exp
+            extended_expression = opt_exp;
+            alloc_mode = Value_mode.regional_to_global_alloc expected_mode.mode
```
lpw25, line 4196: It might be clearer to have the `alloc_mode` be optional here, since the mode for an unboxed record containing regional values will be misleading. That's because the unboxed record is being "allocated" as a regional value in that case -- which can't be expressed as an `Alloc_mode.t`.

typing/typecore.ml
```diff
@@ -4233,10 +4228,11 @@ and type_expect_
         end;
       let mode = mode_cross env ty_arg mode in
       ruem ~mode ~expected_mode {
-        exp_desc = Texp_field(record, lid, label);
+        exp_desc = Texp_field(record, lid, label,
+            Value_mode.regional_to_global_alloc expected_mode.mode
```
lpw25, line 4232: It is probably worth checking the representation of the field and registering the allocation if it is a float field here. Currently we are not always locally allocating floats when we could. It might also be worth making this field optional -- although unlike the `Texp_record` case the value of `alloc_mode` should never be misleading for a `Texp_field` even when it isn't needed.

typing/typecore.ml
```diff
@@ -4251,14 +4247,15 @@ and type_expect_
       if label.lbl_mut = Immutable then
         raise(Error(loc, env, Label_not_mutable lid.txt));
       rue {
-        exp_desc = Texp_setfield(record, label_loc, label, newval);
+        exp_desc = Texp_setfield(record,
+          Value_mode.regional_to_global_alloc rmode,
```
lpw25, line 4251: I think this is a bug. The mode on `Texp_setfield` is not really an allocation mode, in that it does not describe where something should be allocated. Instead it describes whether the record being written to might be on the stack. If it is on the stack then we need to do a slightly different write barrier. In particular, that means that a regional record must be marked as "possibly on the stack", whilst this code is marking it "definitely on the heap".

It would probably be a good idea to just use a different type for this data to try and avoid making this mistake again in future. (For the third time, since this was also a bug in earlier versions of the compiler).

typing/typecore.ml
```diff
@@ -4585,10 +4574,11 @@ and type_expect_
             assert false
       in
       rue {
-        exp_desc = Texp_send(obj, meth, ap_pos);
+        exp_desc = Texp_send(obj, meth, ap_pos,
+          Value_mode.regional_to_global_alloc expected_mode.mode
```
lpw25, line 4578: It would probably be slightly better to register this potential allocation.

typing/typecore.ml
```diff
@@ -6044,7 +6022,8 @@ and type_construct env (expected_mode : expected_mode) loc lid sarg
     end;
   (* NOTE: shouldn't we call "re" on this final expression? -- AF *)
   { texp with
-    exp_desc = Texp_construct(lid, constr, args) }
+    exp_desc = Texp_construct(lid, constr, args,
+      Value_mode.regional_to_global_alloc expected_mode.mode) }
```
lpw25, line 6026: As with the `Texp_record` case it would probably be better for this mode to be optional so that the unboxed case makes sense. That would also allow to leave it as `None` for immediate constructors.

lpw25 (commented):
This looks good, but there are some small improvements that could be made and one bug that needs to be fixed. Probably best to submit those as a single follow up PR.

typing/typecore.ml
riaqn: Indeed it is a bug - originally it was `regional_to_local_alloc`. Yes I will define another type, named either "value_mode" or "access_mode" - which one is better?

typing/typecore.ml
lpw25: I assume so, since there isn't an associated alloc mode in the typedtree.

typing/typecore.ml
lpw25: Maybe "write_barrier_mode" or "modify_mode"?

typing/typecore.ml
lpw25: Yeah

typing/typecore.ml
riaqn: Do you mean to have a new type to replace `alloc_mode` in `Texp_setfield`, or in `Psetfield`? I.e. what stage should this new type be added to?

PR #101, opened 2023-01-18
Fix misplaced attribute warning when using external parser (and some cleanup)

ccasin:
(This replaces #98)

Not all parse trees are created by the parser in the compiler.  There are at least two other possibilities (see [driver/pparse.ml]): 1) the compiler accepts serialized parse trees constructed by other tools, and 2) the compiler can apply rewriters that modify parse trees.  The previous rework of the misplaced attributes check (#44) missed all attributes in case 1 and rewriter-created attributes in case 2.

The first commit in this PR fixes this situation, using the ast invariant checker used by the compiler in cases 1 and 2 to also insert attributes into the table to be tracked for warning 53.  Compared with #98, we don't attempt to give different warnings for rewriter-inserted attributes - this is slightly trickier than I thought because of case 1, and I'd like to get this out quickly, but we can come back later.

The second commit is some minor related cleanup.

### DIFF
--- a/lambda/translattribute.ml
+++ b/lambda/translattribute.ml
@@ -260,12 +260,7 @@ let get_property_attribute l p =
   let attr = find_attribute (is_property_attribute p) l in
   parse_property_attribute attr p
 
-let get_check_attribute l =
-  List.filter_map (fun p ->
-    match get_property_attribute l p with
-    | Default_check -> None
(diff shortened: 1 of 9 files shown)


goldfirere (approved):
Other than translattribute (which I did not review), this looks good to me. I believe @gretay-js did translattribute.

PR #102, opened 2023-01-19
Further tidying around type variables

goldfirere:
In responding to Gabriel's [comment](https://github.com/ocaml/ocaml/pull/11912#discussion_r1080943382) on the upstream PR, I realized that there was an unnecessary call to [instance]. But instead of just removing the call, I added comments and checks to make sure my intuition was correct.

This PR also includes a renaming of `narrow_in` to become `with_local_scope`, which is what upstream uses. (Or will, once my PR there is merged.)

### DIFF
--- a/typing/typetexp.ml
+++ b/typing/typetexp.ml
@@ -75,6 +75,10 @@ module TyVarEnv : sig
   val check_poly_univars : Env.t -> Location.t -> poly_univars -> type_expr list
   (* see mli file *)
 
+  val instance_poly_univars :
+     Env.t -> Location.t -> poly_univars -> type_expr list
+  (* see mli file *)
+
   type policy
   val fixed_policy : policy (* no wildcards allowed *)
   val extensible_policy : policy (* common case *)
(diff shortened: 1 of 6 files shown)


typing/typetexp.ml
```diff
@@ -112,10 +119,42 @@ end = struct
...
+  (* These are variables we expect to become univars (they were introduced with
+     e.g. ['a .]), but we need to make sure they don't unify first.  Why not
+     just birth them as univars? Because they might successfully unify with a
+     row variable in the ['a. < m : ty; .. > as 'a] idiom.  They are like the
+     [used_variables], but will not be globalized in [globalize_used_variables].
+
+     INVARIANT: No types at generic_level.
+  *)
+  let univars = ref ([] : (string * type_expr) list)
+  let assert_univars uvs =
```
stedolan, line 148: Thanks to your comment, I understand that `assert_univars` is the right thing to call before adding to `univars`, but I still find the name `assert_univars` confusing because the name is so unlike the implementation. Maybe `assert_not_generic`?

stedolan (changes requested):

goldfirere:
The story continues at #121.

PR #104, opened 2023-01-23
Refactor toplevel local escape check

stedolan:
In normal compilation units, the following module-level declaration is allowed:

    let _ = local_ ref 42

Module-level bindings must be global, but since nothing is actually bound here it's fine.

At the interactive toplevel, though, it's not fine, because if you do the above then the toplevel prints a line like:

    - : int ref = {contents = 42}

by fishing the value that wasn't bound directly out of the typedtree. If that value is local, then the native toplevel can start segfaulting (See #2).

The current check bans such uses in the toplevel by doing an additional escape check after the fact, but the way it's done is awkward for @riaqn's work in #100. So, this patch refactors that check to instead specify the correct mode *beforehand*, which is both cleaner and does not interfere with #100.


riaqn:
I think it looks good - can we merge this?

(also I guess in the future with more modes axes I will rename `force_global` to `force_toplevel`.)

lpw25 (approved):

PR #105, opened 2023-01-23
Remove pat_mode from typedtree

stedolan:
This patch follows on from @riaqn's #100, removing `pat_mode` from patterns for the same reasons that he removed `exp_mode` from expressions. (This one is a bit simpler because `pat_mode` is used much less)


typing/typedtree.mli
```diff
@@ -76,10 +75,10 @@ and 'k pattern_desc =
   (* value patterns *)
   | Tpat_any : value pattern_desc
         (** _ *)
-  | Tpat_var : Ident.t * string loc -> value pattern_desc
+  | Tpat_var : Ident.t * string loc * Types.value_mode -> value pattern_desc
         (** x *)
   | Tpat_alias :
-      value general_pattern * Ident.t * string loc -> value pattern_desc
+      value general_pattern * Ident.t * string loc * Types.value_mode -> value pattern_desc
```
riaqn, line 81: Quick question: In general I kind of "understand" why we keep the mode for patterns variables and aliases - however, it seems that they are not technically needed? I don't see them being used anywhere.

stedolan: These are needed, they're extracted by `let_bound_idents_with_modes` (some of whose callers want the value modes)

riaqn: I see - thanks!

lambda/translcore.ml
```diff
@@ -237,7 +237,7 @@ let rec push_defaults loc bindings use_lhs cases partial warnings =
                  desc, Id_value)},
              cases, partial) }
       in
-      [{c_lhs = {pat with pat_desc = Tpat_var (param, mknoloc name)};
+      [{c_lhs = {pat with pat_desc = Tpat_var (param, mknoloc name, Value_mode.global)};
         c_guard = None; c_rhs= wrap_bindings bindings exp}]
```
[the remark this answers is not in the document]
stedolan: I was unhappy with this so I've pushed another commit to plumb arg_mode through typedtree, so that we can choose a correct mode here. (This will go away with n-ary lambdas)

stedolan:
This is rebased and I think ready to merge. @riaqn mind having another look?

typing/typecore.ml
```diff
@@ -914,18 +912,18 @@ and build_as_type_aux ~refine (env : Env.t ref) p =
...
 (* Constraint solving during typing of patterns *)
 
-let solve_Ppat_alias ~refine env pat =
+let solve_Ppat_alias ~refine ~mode env pat =
   begin_def ();
-  let ty_var, mode = build_as_type_and_mode ~refine env pat in
+  let ty_var, mode = build_as_type_and_mode ~refine ~mode env pat in
   end_def ();
   generalize ty_var;
   ty_var, mode
```
riaqn, line 929: maybe this function does not need to return `mode` anymore, just to simplify the code a bit?

riaqn (approved):
looks good to me - some minor points.

typing/typecore.ml
riaqn: my bad - the mode could be updated.

PR #107, opened 2023-01-25
Refactoring and fixes around module lookup

rleshchinskiy:
This PR contains refactoring and fixes which improve the code around module type lookup and strengthening. While (hopefully) sensible in their own right, the main driver for this work is improving the performance of module type checking in upcoming PRs which rely on the changes in this PR.

### DIFF
--- a/typing/env.ml
+++ b/typing/env.ml
@@ -1089,7 +1089,7 @@ and find_functor_components path env =
   | Functor_comps f -> f
   | Structure_comps _ -> raise Not_found
 
-let find_module ?(alias=false) path env =
+let find_module path env =
   match path with
   | Pident id ->
       let data = find_ident_module id env in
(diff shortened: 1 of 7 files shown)


typing/env.ml
```diff
@@ -1089,7 +1089,7 @@ and find_functor_components path env =
   | Functor_comps f -> f
   | Structure_comps _ -> raise Not_found
 
-let find_module ~alias path env =
+let find_module ?(alias=false) path env =
```
lpw25, line 1092: I think that this `alias` parameter is unused and can just be removed.

typing/env.ml
```diff
@@ -1103,7 +1103,7 @@ let find_module ~alias path env =
       if alias then md (fc.fcomp_res)
       else md (modtype_of_functor_appl fc p1 p2)
 
-let find_module_lazy ~alias path env =
+let find_module_lazy ?(alias=false) path env =
```
lpw25, line 1106: I think that we do not need to expose this parameter in the interface and it can be left non-optional.

The key observation is that `find_strengthened_module` is the only use of `~alias:true` outside of `Env` after your refactoring, but it actually should be `~alias:false`. Since `alias` only affects the `Papply` case and `find_strengthened_module` is never called with a `Papply` path the difference is not currently detectable, but if it were called with a `Papply` path it would currently give the incorrect answer.

lpw25 (approved):
One small suggestion then this is good to merge.

PR #108, opened 2023-01-26
Constrain curry modes to increase along applications

stedolan:
In short, this patch rejects certain sound-but-weird typings of applications, so that the principal type of an application is less weird. This matters because in certain contexts (particularly, files with no `mli`), the user actually sees the principal type of expressions, so making it not be weird is good.

Specifically, this patch enforces that in a multiple argument application `f a b c d`, then if any of the intermediate partial applications `f`, `f a`, `f a b` and `f a b c` are local, then all subsequent ones must be also.

This does mean that typing can reject more programs than before. In particular, overapplications of functions with local arguments now yield an error:
```
let f (g : local_ string -> int -> (int -> int -> unit)) = g (local_ "a") 2 3 4 [@nontail]
```
yields:
```
Error: This application involving locals is complete after this argument,
       but extra arguments were provided
  Hint: Try wrapping the application in parentheses up to this argument
```
pointing to the second argument `2`.

As the hint suggests, replacing the body with `(g (local_ "a") 2) 3 4` (splitting the overapplication into two applications) resolves the issue.

The benefit is that if the user writes the same function without a type annotation, as:
```
let f g = g (local_ "a") 2 3 4 [@nontail]
```
then in a file with no mli, this patch changes the inferred type to:
```
val f : (g : local_ string -> int -> int -> int -> 'a) -> 'a
```
which is probably what the user intended.

### DIFF
--- a/testsuite/tests/typing-local/local.ml
+++ b/testsuite/tests/typing-local/local.ml
@@ -326,12 +326,12 @@ val apply2 : int -> local_ 'a -> int -> int = <fun>
 |}]
 let apply3 x = g x x x
 [%%expect{|
-Line 1, characters 19-20:
+Line 1, characters 15-20:
 1 | let apply3 x = g x x x
-                       ^
-Error: This application involving locals is complete after this argument,
-       but extra arguments were provided
-  Hint: Try wrapping the application in parentheses up to this argument
+                   ^^^^^
+Error: This application is complete, but further arguments were provided afterwards.
+       With local arguments or closures, these are not allowed in the same application.
+  Hint: Try wrapping the marked application in parentheses.
 |}]
 let apply3_wrapped x = (g x x) x
 [%%expect{|
@@ -345,12 +345,12 @@ Error: This local value escapes its region
 |}]
 let apply4 x = g x x x x
 [%%expect{|
-Line 1, characters 19-20:
+Line 1, characters 15-20:
 1 | let apply4 x = g x x x x
-                       ^
-Error: This application involving locals is complete after this argument,
-       but extra arguments were provided
-  Hint: Try wrapping the application in parentheses up to this argument
+                   ^^^^^
+Error: This application is complete, but further arguments were provided afterwards.
+       With local arguments or closures, these are not allowed in the same application.
+  Hint: Try wrapping the marked application in parentheses.
 |}]
 let apply4_wrapped x = (g x x) x x
 [%%expect{|
@@ -420,12 +420,12 @@ Error: This local value escapes its region
 let app42 (f : a:local_ int ref -> (int -> b:local_ int ref -> c:int -> unit)) =
   f ~a:(local_ ref 1) 2 ~c:4
 [%%expect{|
-Line 2, characters 7-21:
+Line 2, characters 2-21:
 2 |   f ~a:(local_ ref 1) 2 ~c:4
-           ^^^^^^^^^^^^^^
-Error: This application involving locals is complete after this argument,
-       but extra arguments were provided
-  Hint: Try wrapping the application in parentheses up to this argument
+      ^^^^^^^^^^^^^^^^^^^
+Error: This application is complete, but further arguments were provided afterwards.
+       With local arguments or closures, these are not allowed in the same application.
+  Hint: Try wrapping the marked application in parentheses.
 |}]
 let app42_wrapped (f : a:local_ int ref -> (int -> b:local_ int ref -> c:int -> unit)) =
   (f ~a:(local_ ref 1)) 2 ~c:4
@@ -483,12 +483,12 @@ Error: This local value escapes its region
 let app42' (f : a:local_ int ref -> (int -> b:local_ int ref -> c:int -> unit)) =
   f ~a:(ref 1) 2 ~c:4
 [%%expect{|
-Line 2, characters 7-14:
+Line 2, characters 2-14:
 2 |   f ~a:(ref 1) 2 ~c:4
-           ^^^^^^^
-Error: This application involving locals is complete after this argument,
-       but extra arguments were provided
-  Hint: Try wrapping the application in parentheses up to this argument
+      ^^^^^^^^^^^^
+Error: This application is complete, but further arguments were provided afterwards.
+       With local arguments or closures, these are not allowed in the same application.
+  Hint: Try wrapping the marked application in parentheses.
 |}]
 let app42'_wrapped (f : a:local_ int ref -> (int -> b:local_ int ref -> c:int -> unit)) =
   (f ~a:(ref 1)) 2 ~c:4
@@ -500,12 +500,12 @@ val app42'_wrapped :
 let app43' (f : a:local_ int ref -> (int -> b:local_ int ref -> c:int -> unit)) =
   f ~a:(ref 1) 2
 [%%expect{|
-Line 2, characters 7-14:
+Line 2, characters 2-14:
 2 |   f ~a:(ref 1) 2
-           ^^^^^^^
-Error: This application involving locals is complete after this argument,
-       but extra arguments were provided
-  Hint: Try wrapping the application in parentheses up to this argument
+      ^^^^^^^^^^^^
+Error: This application is complete, but further arguments were provided afterwards.
+       With local arguments or closures, these are not allowed in the same application.
+  Hint: Try wrapping the marked application in parentheses.
 |}]
 let app43'_wrapped (f : a:local_ int ref -> (int -> b:local_ int ref -> c:int -> unit)) =
   (f ~a:(ref 1)) 2
--- a/testsuite/tests/typing-local/local.ml
+++ b/testsuite/tests/typing-local/local.ml
@@ -556,7 +556,13 @@ let bug2 () =
   let res = bar ~a:"world" in
   res
 [%%expect{|
-val bug2 : unit -> c:int -> unit = <fun>
+Line 5, characters 19-33:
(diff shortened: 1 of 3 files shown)


testsuite/tests/typing-local/local.ml
```diff
@@ -320,15 +320,57 @@ val apply2 : int -> local_ 'a -> int -> int = <fun>
...
 |}]
 let apply4 x = g x x x x
 [%%expect{|
-val apply4 : int -> int = <fun>
+Line 1, characters 19-20:
+1 | let apply4 x = g x x x x
+                       ^
+Error: This application involving locals is complete after this argument,
+       but extra arguments were provided
+  Hint: Try wrapping the application in parentheses up to this argument
```
goldfirere, line 345: I think a little time iterating on the error message may be time well spent. Here is my stab:

```suggestion
Line 1, characters 19-20:
1 | let apply4 x = g x x x x
                   ^^^^^
Error: The function application this message points to is complete,
       but extra arguments were provided.
       This is not allowed with local arguments.
  Hint: Try wrapping this application in parentheses
```

The big change here is to label the entire application, not just the last argument. (Not sure how hard that is in practice, but hopefully not too bad.) It took me a little while to understand "this argument".

I'm still not thrilled with my message, but maybe it inspires you do continue to improve.

goldfirere (commented):
I did not review the implementation yet -- focusing still on the error message.

Does this change handle the "your function that has optional arguments actually doesn't take as many arguments at once as you think" problem? I suppose not, because that doesn't interact with functions not returning locals. Too bad. But this is still a nice change.

testsuite/tests/typing-local/local.ml
stedolan: Yes, that does look better. It's mildly annoying to implement because we don't have that location to hand (there is no corresponding parsetree node), but the improvement seems worth it.

stedolan: This was easier than I expected! Better error message now pushed.

testsuite/tests/typing-local/local.ml
```diff
@@ -320,15 +320,57 @@ val apply2 : int -> local_ 'a -> int -> int = <fun>
...
 let apply3 x = g x x x
 [%%expect{|
-Line 1, characters 15-22:
+Line 1, characters 15-20:
 1 | let apply3 x = g x x x
-                   ^^^^^^^
+                   ^^^^^
+Error: This application is complete, but further arguments were provided afterwards.
+       With local arguments or closures, these are not allowed in the same application.
+  Hint: Try wrapping the marked application in parentheses.
```
goldfirere, line 328: Nitpick: I honestly read "these" as referring first to "closures", which didn't make sense, and then "local arguments or closures", which still didn't make sense, and then I finally found "further arguments". With that in mind, what do you think of:

```suggestion
Error: This application is complete, but further arguments were provided afterwards.
       When passing local values, extra arguments must appear in a separate application.
  Hint: Try wrapping the marked application in parentheses.
```

typing/typecore.ml
```diff
@@ -2714,7 +2715,27 @@ let remaining_function_type ty_ret mode_ret rev_args =
   in
   ty_ret
 
-let collect_unknown_apply_args env funct ty_fun mode_fun rev_args sargs ret_tvar =
+let check_local_application_complete ~env ~app_loc ~sarg ~rest ~mode_fun ~mode_arg ~mode_ret =
```
goldfirere, line 2718: May I order a comment, please? Here is a start:

```
(* This function checks whether an application to local arguments is complete.
   It is necessary to avoid weird types being inferred in places where e.g. there is
   no mli file. For example, TODO should get the type TODO, not TODO.
   The error this produces is somewhat obscure, so prefer other errors over this one
   when there is a choice. This check could be dropped -- there's no soundness issue here --
   at the cost of a worse user interface. *)
```

I worry that, without such a comment, we'll be scratching our heads over this check in a few years.

typing/typecore.ml
```diff
@@ -2741,14 +2762,18 @@ let collect_unknown_apply_args env funct ty_fun mode_fun rev_args sargs ret_tvar
...
+              let kind = (lbl, mode_arg, mode_ret) in
               unify env ty_fun
                 (newty (Tarrow(kind,ty_arg,ty_res,commu_var ())));
-              (mode_arg, ty_arg_mono, mode_res, ty_res)
-        | Tarrow ((l, mode_arg, mode_res), ty_arg, ty_res, _)
+              (mode_arg, ty_arg_mono, mode_ret, ty_res)
+        | Tarrow ((l, mode_arg, mode_ret), ty_arg, ty_res, _)
           when labels_match ~param:l ~arg:lbl ->
-            (mode_arg, tpoly_get_mono ty_arg, mode_res, ty_res)
+            check_local_application_complete ~env ~app_loc ~sarg ~rest
```
goldfirere, line 2774: Why don't the checks here need to be delayed like elsewhere?

stedolan: In this case, we're looking at an application to an unknown function and inferring the function's type, and it wasn't obvious to me that the default order of checks was wrong. (In the other case, we *know* the type of the function, and we should really use that known information before we start messing about with modes).

(Having said that, I've no problem with delaying the checks here too. I don't feel strongly either way so I did the thing that took the least code)

testsuite/tests/typing-local/local.ml
goldfirere:
```
Error: This application is complete, but further arguments were provided afterwards.
       When passing or calling a local value, extra arguments are passed in a separate application.
  Hint: Try wrapping the marked application in parentheses.
```

?

typing/typecore.ml
goldfirere: Make sense to me.

typing/typecore.ml
```diff
@@ -2778,6 +2804,18 @@ let collect_apply_args env funct ignore_labels ty_fun ty_fun0 mode_fun sargs ret
...
+        | (_lbl, sarg) :: rest ->
+          (* This is an obscure error, which often triggers on ill-typed applications.
+             Since it is relatively confusing, we want to delay issuing it until we're
+             sure the application is otherwise well-typed, because in ill-typed cases
+             the standard error is easier to understand *)
+          delayed_checks :=
+            (fun () -> check_local_application_complete ~env ~app_loc ~sarg ~rest
+                         ~mode_fun ~mode_arg ~mode_ret)
+            :: !delayed_checks
+        end;
```
lpw25, line 2818: I think it would be better to avoid mixing this check up with the complicated argument gathering logic. The result of `collect_apply_args` already has most of the needed information for the check. I think it is just missing `mode_fun` in the `Arg (Known_arg ...)` and `Arg (Unknown_arg ...)` cases, and it's readily available at the point where those constructors are created. So rather than a higher-order ref, you can just iterate over that data at the point where you want to do the check (it's currently shadowed by the output of `type_apply_arg` but that's easy to fix).

typing/typecore.ml
```diff
@@ -2714,7 +2715,27 @@ let remaining_function_type ty_ret mode_ret rev_args =
...
-let collect_unknown_apply_args env funct ty_fun mode_fun rev_args sargs ret_tvar =
+let check_local_application_complete ~env ~app_loc ~sarg ~rest ~mode_fun ~mode_arg ~mode_ret =
+  let submode m1 m2 =
+    match Alloc_mode.submode m1 m2 with
+    | Ok () -> ()
+    | Error () ->
+      let loc =
+        Location.{ loc_start = app_loc.loc_start;
+                   loc_end = sarg.pexp_loc.loc_end;
+                   loc_ghost = app_loc.loc_ghost || sarg.pexp_loc.loc_ghost }
```
lpw25, line 2726: I think this location may be confusing in the case where the arguments have been passed out of order. I think that sometimes there will be no single span that corresponds to the complete application, so we'll need to think of something else to do there.

lpw25 (commented):
I think that we should slightly move where the check is taking place, although I believe the current code to be mostly correct.

typing/typecore.ml
lpw25: Sure, but I think this check goes away because we will be assuming that an application to n arguments is for an n parameter function, which will already force the desired property here. (We will still allow over-application when the function type is know, but that is also fine from this perspective).

This check is essentially adding mode constraints to enforce "if the function or a parameter is local then function should be at least arity n", whereas later we will just always enforce that the arity regardless of the modes of anything.

testsuite/tests/typing-local/local.ml
```diff
@@ -571,6 +577,52 @@ Line 3, characters 63-64:
...
+
+let () = overapp ~b:2 ~a:1 ~c:3 ~d:4
+[%%expect{|
+Line 1, characters 20-21:
+1 | let () = overapp ~b:2 ~a:1 ~c:3 ~d:4
+                        ^
+Error: This application is complete, but surplus arguments were provided afterwards.
+       When passing or calling a local value, extra arguments are passed in a separate application.
+  Hint: Try splitting the application in two, first applying the arguments
+  up to the marked one (in the function's type), and then the rest.
```
lpw25, line 601: The wording on this hint is particularly awkward. How about:
```
  Hint: Try splitting the function application in two. The arguments that come after this
  one in the function's type should be applied separately.
```

lpw25 (approved):
I think the wording on one hint might be better, but this LGTM.

testsuite/tests/typing-local/local.ml
stedolan: Yeah, that's obviously much better.

PR #111, opened 2023-01-27
Consistently use Lmutvar or Lvar in comprehensions

goldfirere:
This commit adds an `occur` field to `Let_binding` so that it can correctly use `Lvar` or `Lmutvar`. Tiny bonus: fewer allocations from Lvar nodes, because now we just create one.

### DIFF
--- a/lambda/transl_array_comprehension.ml
+++ b/lambda/transl_array_comprehension.ml
@@ -228,16 +228,17 @@ end = struct
       if this overflowed; the inputs are required to be values, as they are
       evaluated more than once *)
   let safe_mul_pos_vals ~loc x y =
+    let y = y.Let_binding.var in
     let open (val Lambda_utils.int_ops ~loc) in
-    let product, product_binding =
-      Let_binding.make_var (Immutable Alias) Pintval "product" (x * y)
+    let product =
+      Let_binding.make (Immutable Alias) Pintval "product" (x * y)
     in
     (* [x * y] is safe, for strictly positive [x] and [y], iff you can undo the
        multiplication: [(x * y)/y = x].  We assume the inputs are values, so we
        don't have to bind them first to avoid extra computation. *)
-    Let_binding.let_one product_binding
-      (Lifthenelse(product / y = x,
-         product,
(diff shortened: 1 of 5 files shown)


antalsz:
This looks great!  Can I bikeshed by suggesting we change `occur` to `var`?  And maybe add a slightly longer comment explaining what it is?

antalsz (approved):

goldfirere:
Well, it's specifically an occurrence of the variable, but `occurrence` was just way too long. Though I agree the fact that I used a verb there is off-putting. `var` it is.

Also, in case you're tempted to do the same: I actually tried to make all of these be GADTy, so that you couldn't actually extract the `id` of a `Let_binding.t` and instead had to use `assign : mutable_ t -> lambda -> lambda` to make an `Lassign`. I'm sure this would work in the end, but GADT type inference is too painful (needed some extra annotations and eta-expansions), and there's no e.g. `Let_binding.(immutable t)` syntax in types. It would all just be too clutterful in the end.

Sidenote: we should add `Let_binding.(immutable t)` syntax. That was the straw that broke my back on this one.

antalsz:
Looks great, merged!

If we want to pursue the GADT route:

1. OCaml can't rely on two empty types being disjoint (this is why you've seen `type index = private Index` before).  I think this is because you could have
```
module M : sig
  type t1 = |
  type t2 = |
  
  val x1 : t1 option
  val x2 : t2 option
end = struct
  type t1 = |
  type t2 = t1 = |
  
  let x1 : t1 option = None
  let x2 : t2 option = None

  let _ = [x1; x2]
end```
and then `M.[x1; x2]` would be a type error.  Abstraction!  It takes some getting used to :-)  Fixing that might help your type inference.

2. I would probably just define `type mutable_t = mutable_ t` and `type immutable_t = immutable t`, or maybe better,
```
type immutable_let = private Immutable_let
type mutable_let   = private Mutable_let

type immutable = immutable_let t
type mutable   = mutable_let t
```
(That said, we should add local opens in types, I think this idea has come up before.)

PR #112, opened 2023-01-31
Print `nlocal` in the `-d(raw)lambda` output

antalsz:
I'm open to bikeshedding on the syntax (we want to be able to read this nicely!), but I think we should print this somewhere; without it, I was lead astray in thinking two things were the same that weren't.


antalsz:
It looks like there's a failing test in CI because we run a different backend there.  I'll fix that, but I dunno how to check if anything else is lurking.

goldfirere:
I nominate @lukemaurer to review.

lukemaurer (approved):
Looks sensible to me

PR #113, opened 2023-01-31
Fix `nlocal` in the generated Lambda for list comprehensions

antalsz:
This fixes the bug that was causing list comprehensions to fail to build with stack allocation enabled, and reenables all the tests


goldfirere (approved):

riaqn (approved):
LGTM

PR #115, opened 2023-02-01
Don't track attributes inside attributes for warning 53

ccasin:
Attributes can occur in the payload of other attributes (or extension nodes).  [This happens in practice](https://github.com/LexiFi/gen_js_api/blob/d91966229284995f179d1889c822e57853f6267e/examples/misc/jquery.mli#L8), if rarely.  With the reworked misplaced attributes system, we end up giving spurious warnings for these inner attributes.  The reason is that we just add every attribute we parse to the table, and the inner attributes are generally not "used" so they are never removed.

This PR fixes this issue in a somewhat heavy handed way: After parsing an attribute or extension payload, we simply traverse it and remove any attributes it contains from the table tracking attributes for the warning.  One alternative would be to parameterize nonterminals in the parser to indicate whether it happens inside an attribute, but this seems really noisy because basically everything would need a parameter.  Another alternative would be to use some kind of state within the parser, but this seems quite error-prone.  It think the current approach is the best compromise, and certainly the easiest to write.

There are a bunch of changes in makefiles because `Builtin_attributes` now depends on `Ast_iterator` and it didn't previously.  This doesn't seem problematic, but perhaps I'm missing something.


goldfirere:
Why are attributes-in-attributes not used? If they're not used, that sounds like they should indeed give rise to a warning. I think an example of how this happens in practice may be helpful.

ccasin:
> Why are attributes-in-attributes not used? If they're not used, that sounds like they should indeed give rise to a warning. I think an example of how this happens in practice may be helpful.

See the example I linked in the first message.  This attribute is used, in that the gen-js-api ppx does something with it (inserts it into a generated ml file).  But that ppx may be run in a separate process so it's hard for that information to be communicated back to us.

In general you could imagine worse situation, like a ppx that does some kind of partial evaluation or metaprogramming implementing a static "if" that throws away half the code in an attribute payload, and that code might have attributes that get parsed and added to our table (depending on exactly when the ppx runs and whether the attribute sticks around). I don't think it's feasible to interpret whether the payloads of arbitrary attributes intended for external tools are used.

ccasin:
I discussed this with Richard, but to quickly document a couple of the answers:

- We can't guarantee that ppxs will remove their attributes and their payloads, so we must handle the case where attributes remain in attribute payloads after the ppx runs.  These will generally not be used by the compiler, but shouldn't result in a warning.
- There are two ways to invoke a ppx: `-pp`, where the external ppx application is called first and hands its output back to the compiler, and `-ppx` where the compiler first parses the file, then runs the ppx, then checks that the resulting parse tree respects various invariants.  It's true that we have a bug where, in the latter case, we might warn about compiler built-in attributes which are removed by the ppx.  Of course, a ppx shouldn't do that, so it doesn't happen in practice, and it's a bit orthogonal to what is fixed in this PR, so I'll fix it separately.

ccasin:
(There do remain a couple issues with this, so hold off on reviewing.  I am fixing them now and will comment again when I'm done).

ccasin:
OK, it took me a minute to rationalize the story here, and I explain it in unfortunate detail below.  It's now ready for review.

Goals for the attributes tracking mechanism:
- Add attributes to the table just once.
- They should be the attributes in the ast that's going to get passed to the typechecker, not any ASTs before preprocessing, regardless of what combination of `-pp` and `-ppx` you use.
- For attributes within attribute payloads: either don't add them to the table, or remove them from the table before typechecking.  Attributes may be left behind by ppxes and we have no way to asses the meaning of their payloads.
- For attributes within extension payloads: do add them to the table. We're only looking at the final version of the parse tree, so any extension nodes handled by ppxes are gone, leaving only extension nodes that the compiler must handle.

I went back and had a more careful look at the way the compiler reads a file.  A thing that was not obvious to me previously is that you can pass both `-pp` and `-ppx`.  They do different things that have related uses.  Here's the rough series of events:

1) If you pass `-pp`, a preprocessor you specify runs on the file before ocaml tries to parse it.

2) We read the file.  This file may be either a normal source file or a marshalled AST (perhaps produced by the `-pp` program, though those can also produce normal source files, and I suppose you could pass ocaml a marshalled ast without `-pp`).
    - If it's a marshalled AST: it's unmarshalled.  If you have not passed the `-ppx` flag, we check parse tree invariants now.
    - If it's a normal ocaml file, it's parsed.

3) Now we have a parse tree.  We check if you have passed `-ppx`:
    - If not, we're done.
    - If so, the parse tree from step 2 is marshalled to disk. That file is passed to your ppx, producing another marshalled ast that the compiler then unmarshalls. We check the ast invariants for this final parse tree.

Observe that on paths where the invariant checker is going to be run, it runs just once and always on the final parse tree.  Also, it's easy for the invariant checker to skip adding attributes inside of other attributes.  So, this seems like the place to add attributes to the table when it's an option.

The only path where the invariant checker does not run is the case where step 2 encounters a normal source file and the `-ppx` flag is not used.  To handle that case, we add attributes to the table during the parsing in step 2.  As described in the description of this PR, we have a slightly hacky solution to subsequently remove attributes in attributes.  Additionally, we need a way to tell the parser _not_ to add attributes to the table in the case where we're parsing a file but a ppx will run later.

(An alternative would be to just always check the ast invariants, even if the final parse tree comes from the parser.  This occurs a linear penalty in the externally-common scenario that we're just parsing a regular source file and have no ppxes, so I suspect it wouldn't fly upstream.)

Based on this understanding, I've pushed a commit that makes the following changes:
- When adding attributes during the invariants check, we are careful not to add attributes-in-attributes.
- When adding attributes during the parser, I have fixed it so that attributes in extensions _are_ added (this PR previously mistakenly removed them).
- At the place where we run the parser in `Pparse`, we check if we're subsequently going to run a ppx and turn off attribute tracking if so.  This required adding some state to `Builtin_attributes`, which I don't love, but it seemed better than threading it all the way through the parser.
- I've added a test to check that we get misplaced attribute warnings when the compiler is given a marshalled ast.  There might be a cleaner way to write this test, but I didn't find a way to generated a marshalled ast other than writing a program to do it with the compilerlibs.

goldfirere:
Thanks for this detailed explanation!

A few reactions before reviewing:
 - I don't think avoiding a linear amount of work is the right goal here. By its nature, this check must do a linear amount of work: it's checking every attribute in the AST. Maybe we should make sure to skip this work if the warning isn't enabled (as opposed to, say, doing the work but squelching the output), but I don't think we should necessarily discard any approach that requires linear extra work (indeed we can't do this!). Of course, doing a whole separate AST pass just for this check may still be problematic, because it has to load all the data, etc., causing cache misses and whatnot. So I'm not really arguing here for a different approach, but just that "linear amount of work" isn't the relevant criterion.
 - The explanation above detailing the reasons behind the design here is great. Can this be persisted in the source code? I'm imagining a

```
Note [Unused attributes check]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Warning 73 (-missing-attributes) checks to make sure all attributes the
compiler itself handles are actually used; an unused one is in a wrong
location and should be flagged to the user.

This warning is implemented via a table that stores TODO. It is
created in TODO and processed in TODO. The actual warning is emitted
in TODO.

We must be careful in the presence of preprocessors. We thus use the
following scheme:
- When `-pp` is passed, TODO.
- When `-ppx` is passed, TODO.

An interesting case is when code appears in payloads in other attributes
or extensions. These cases are actually handled differently:
- In attributes, TODO. The relevant code appears at TODO.
- In extensions, TODO. The relevant code appears at TODO.
```

and then all the code referenced in the Note or other code dealing with this warning can have `See Note [Unused attributes check]`.

In the end, you've basically written the note contents already, but if it is in the source code itself, then it is incomparably more discoverable than sitting in a comment in this PR. And if it's useful to us now, it will be useful to all OCaml contributors in the future.

- Should a new entry be added to the OCaml manual? This is actually a larger question than just this PR, and I will duly raise it elsewhere.

parsing/builtin_attributes.ml
```diff
@@ -92,15 +92,17 @@ let builtin_attrs =
 
 let is_builtin_attr s = Hashtbl.mem builtin_attrs s
 
+let track = ref true
```
goldfirere, line 95: Right around here would be a great spot for that Note.

Even if we decide not to do the full glorious Note (but I think we should!), please comment this state.

parsing/builtin_attributes.ml
```diff
@@ -92,15 +92,17 @@ let builtin_attrs =
...
+let track_attrs b = track := b
```
goldfirere, line 96: What about

```ocaml
val track_attrs_in : track_attrs:bool -> (unit -> 'a) -> 'a
```

that uses `protect_refs`? Seems considerably safer and easier to reason about then the current very stateful approach. I'm not really objecting to the state, per se, but just the interface to interacting with it.

parsing/builtin_attributes.mli
```diff
@@ -40,12 +40,17 @@
...
-    than Attr.mk for all built-in attributes. *)
+    are tracked for misplaced attribute warnings (unless this tracking has been
+    turned off with [track_attrs]).  This should be used rather than Attr.mk for
+    all built-in attributes. *)
 val mk_internal:
   ?loc:Location.t -> string Location.loc -> Parsetree.payload ->
   Parsetree.attribute
 
+(** Turns on or off the tracking of attributes constructed with [mk_internal]
+    for warning 53. *)
```
goldfirere, line 51:
```suggestion
    for warning 53. Default is "on". *)
```

parsing/builtin_attributes.mli
```diff
@@ -40,12 +40,17 @@
 
 
 (** Used by the compiler for attributes that may be compiler built-ins.  These
-    are tracked for misplaced attribute warnings.  This should be used rather
-    than Attr.mk for all built-in attributes. *)
+    are tracked for misplaced attribute warnings (unless this tracking has been
+    turned off with [track_attrs]).  This should be used rather than Attr.mk for
```
goldfirere, line 44: The last sentence here (not new in this patch) surprises me. If I'm parsing an attribute `f x y [@blah]`, then should I really check whether `blah` is built-in to decide whether to call `Attr.mk` or `mk_internal`?

parsing/builtin_attributes.mli
```diff
@@ -61,6 +66,10 @@ val mark_alerts_used : Parsetree.attributes -> unit
     attributes into the environment. *)
 val mark_warn_on_literal_pattern_used : Parsetree.attributes -> unit
 
+(** Marks the attributes hiding in the payload of another attribute (or
+    extension node) used, for the purposes of misplaced attribute warnings. *)
```
goldfirere, line 70: A natural question is: why have this function? The answer would be in the Note. :grin:

parsing/parser.mly
```diff
@@ -1067,6 +1067,8 @@ The precedences must be listed from low to high.
 %inline mk_directive_arg(symb): symb
     { mk_directive_arg ~loc:$sloc $1 }
 
+%inline mark_payload_attrs_used(symb): symb
```
goldfirere, line 1070: This seems unnecessarily indirect. Just call `Builtin_attributes.mark_payload_attrs_used` in the body of `attr_payload`. No need for this parameterized parser non-terminal.

Actually, even better: could you use the new stateful approach to avoid adding to the table at all when in an attribute payload? I think you could, and that would save you the extra walk over the attribute payload in `mark_payload_attrs_used`.

ccasin:
To be more precise about the extra work thing: The current approach requires extra work that's linear in the total size of attribute payloads (generally very small), while the other would require work linear in the size of the parse tree.   Still not necessarily a big deal, but not identical.

parsing/parser.mly
ccasin: I didn't take the latter approach because I was trying to be a bit defensive: I don't fully understand the code menhir generates and was concerned about what might happen to the state in error cases.

ccasin: (But I guess if we use protect refs and put it just in the action, it's fine.  I'll make that change.)

ccasin: Sorry - I think that last comment is wrong.  So to explain in more detail why I didn't take the latter approach: It seems tricky - by the time we're in the action of the attribute nonterminal, it's too late to change the state.  So we need some more complicated set up with additional non terminals that guard the parsing of attributes.  This, combined with the fact that the action of those nonterminals would be to change state, seemed possible but hard to get right.

ccasin: (And as for the `mark_payload_attrs_used` bit - my impression was that this approach is cleaner and more idiomatic, but as you disagree I'll move it.)

lpw25:
This is a meta point -- and I don't want to side-track things too much with it -- but I remain somewhat unconvinced by large stand-alone comments like that. I mostly think of them as a code smell because people tend to write them when they should instead make the code clearer. Can we do something to make the logic from that comment clearer in the code itself?

My disagreement here isn't that the suggested information shouldn't be in the source -- it definitely should -- but I think it can be better done by designing a good API and commenting the elements of that, rather than including a big prose discussion.

goldfirere:
> This is a meta point -- and I don't want to side-track things too much with it -- but I remain somewhat unconvinced by large stand-alone comments like that. I mostly think of them as a code smell because people tend to write them when they should instead make the code clearer. Can we do something to make the logic from that comment clearer in the code itself?
> 
> My disagreement here isn't that the suggested information shouldn't be in the source -- it definitely should -- but I think it can be better done by designing a good API and commenting the elements of that, rather than including a big prose discussion.

I'm all for designing a good API that makes these facts emergently apparent. That's definitely better than the prose comment, for many reasons. The problem is that something like this warning is implemented as small snippets of code spread throughout a number of places, and these small snippets will surely struggle to create a whole-picture view without a prose comment. Or maybe let me put this another way: an alternative approach to the implementation of this warning is to have a whole new pass, right before the type-checker, that does everything. The source code for that pass could be carefully written to contain much (all?) of the content in my imagined comment. That would be easier to understand, maintain, and reason about. But it would be slower! So the version that Chris has implemented is essentially an optimized version of the one big pass, and, as usual, the optimized version is harder to understand. The prose comment makes up for that loss and can be seen as a fragment of the optimization, rather than something inherent to the approach. So, yes, the comment smells a bit... like the burnt rubber made by going fast.

lpw25:
> The problem is that something like this warning is implemented as small snippets of code spread throughout a number of places, and these small snippets will surely struggle to create a whole-picture view without a prose comment.

This certainly makes it harder to keep all the logic in one place but it doesn't make it impossible. You can still basically always write a single API that encapsulates all the logic with separate entry points for the various places that interact with this logic. The current PR doesn't really do that -- `Builtin_attributes` exposes fairly low-level operations and expects the rest of the code to know how to use them. For example, it has logic for turning checking on and off in various places, instead it could just always call into `Builtin_attributes` and provide enough information at the call sites to decide whether to actually do anything.

goldfirere:
Yes, that's a good point -- all the distributed bits and pieces could be function calls into one module. Then my free-floating comment becomes a module comment... and it really might end up just as a bunch of comments on individual functions, with (ideally) nothing left over. This is better than a big prose comment, indeed.

ccasin:
> > The problem is that something like this warning is implemented as small snippets of code spread throughout a number of places, and these small snippets will surely struggle to create a whole-picture view without a prose comment.
> 
> This certainly makes it harder to keep all the logic in one place but it doesn't make it impossible. You can still basically always write a single API that encapsulates all the logic with separate entry points for the various places that interact with this logic. The current PR doesn't really do that -- `Builtin_attributes` exposes fairly low-level operations and expects the rest of the code to know how to use them. For example, it has logic for turning checking on and off in various places, instead it could just always call into `Builtin_attributes` and provide enough information at the call sites to decide whether to actually do anything.

I'm confused by this remark.  The reason we have the function to turn on and off adding things to the table is exactly that we don't have enough information at the call sites of the other relevant Builtin_attributes function to decide whether the attribute in question should be added.  At least, not without threading information through every parser production in a way that seems undesirable.

lpw25:
Isn't the relevant information a combination of "this attribute is coming from the parser" and the various Clflag stuff to determine whether or not we are adding attributes from the parser. Is was picturing something like the following but with better names:
```ocaml
type where_we_are_adding_stuff =
  | Parsing
  | Invariant_check
  
let choose_where_we_are_adding_stuff () =
    match !Clflags.ppx, !Clflags.pp with
    | [], [] -> Parsing
    | _ :: _, _ | _, _ :: _ -> Invariant_check

let where_we_are_adding_stuff_state = ref None

let where_we_are_adding_stuff () =
  match !where_we_are_adding_stuff_state with
  | Some where -> where
  | None ->
      let where = choose_where_we_are_adding_stuff () in
      where_we_are_adding_stuff_state := Some where;
      where
      
let mk_from_parsing attr =
  match where_we_are_adding_stuff () with
  | Parsing -> add_the_attr attr
  | Invariant_check -> ()
```
Possibly `where_we_are_adding_stuff_state` isn't needed and you can just work it out each time an attribute is encountered.

lpw25:
If you don't want to read the state of `Clflags` directly you could also have `pparse` pass it through a function. But either way there is no need to directly expose the ability to enable and disable tracking of attributes.

ccasin:
Yes, good point.  (I had mixed up the check to skip adding because we'll run the invariant check later with the attributes-in-attributes thing, in my head.)

I'll implement something like your api.

lpw25:
FWIW I'd also be tempted to slightly adjust the attributes-in-attributes thing to be something like:
```ocaml
type is_attribute_nested =
  | Unnested_attribute
  | Nested_attribute
  
val register_attr: is_attribute_nested -> string Location.loc -> unit
```
where `register_attr Nested_attribute` is a no-op. Then the code in the invariants check is tracking whether attributes are nested and relaying that information, rather than explicitly turning off the check.

ccasin:
I pushed a commit that I think addresses this discussion and improves the API.

I deleted `Builtin_attributes.mk_internal`: It was a nice thought that we could hide the idea of "registering" attributes behind the interface for their creation, but this just doesn't make any sense for attributes materialized from marshalled ASTs.  So it seems better to just have one path in the API for attributes to get added to the table - calling `Builtin_atributes.register_attr` explicitly in both the parser case and the `Ast_invariants` case.

`register_attr` now takes a parameter indicating whether it's being called from the parser or the invariant checker.  It directly reads the clflags, and this eliminates the need to add more state.  Attributes are always created with `Attr.mk`, as they used to be, and then we call `register_attr` if appropriate.

I opted not to add a `is_nested_attribute`, as Leo suggested in his last comment, because this parameter can't be accurately filled in when `register_attr` is called from the parser.  We could simply pretend all parsed attributes are not nested, but this seems misleading.

I've expanded the comment on `register_attr`, which is now most of the API, but let me know if you'd like more.

goldfirere (approved):
Just a few comment clarifications requested. I agree that this doesn't need a big summary comment any more. I optimistically Approve; feel free to merge when you're ready.

PR #116, opened 2023-02-03
add `unique` extension

riaqn:
This PR is based on the previous PR (https://github.com/ocaml-flambda/ocaml-jst/pull/33), and tries to add support for the uniqueness extension.
- Add two new axes to the mode system. The `uniqueness` axis contains `unique` < `shared`. The `linearity` axis contains `many` < `once`.  To accomodate the new axes, some refactoring and renaming are done. More of that is due but I think it's better to merge this ASAP instead of rebasing weekily. 
- To check that variable uses declared to be `unique` are actually so, we add `uniqueness_analysis.ml` that checks the usage of each values (In particular, `Texp_ident` and `Texp_field`).

The extension still has many issues, our minimal goal in this PR is that existing code will translate to the same machine code when the `unique` extension is disabled. Existing code might be rejected by type checker due to some bugs we introduced, but that's less scary. To achieve this minimal goal, the following is needed:

- We first make sure that this PR doesn't change the `locality` axis in the `alloc_mode` in the `typedtree` generated by `typecore.ml`.
- In `translcore.ml`, add assertions that the two new lambda primitives (that overwrite fields in blocks), `Preuseblock` and `Preusefloatblock` are generated _only_ when `unique` extension enabled.
- Note that in `lambda.ml`, we add `uniqueness` to `alloc_mode`. `uniqueness` could be either `alloc_unique` or `alloc_shared`. The former is redirected to the latter when the extension is disabled. Further inspection is needed to ensure that `translcore.ml` retains old behaviour when `uniqueness` is `alloc_shared`.


riaqn:
This is closed in favour of https://github.com/ocaml-flambda/ocaml-jst/pull/138

PR #117, opened 2023-02-06
Correctly handle unclosed immutable array patterns

antalsz:
While I made the `[| ... |]` and `[: ... :]` expression forms uniform, I didn't make the pattern forms uniform, and I missed one case which gave improved error messages.  This commit rectifies that by bringing both the expression and pattern syntax under the `Generic_array` umbrella.


goldfirere (approved):

parsing/parser.mly
```diff
@@ -310,26 +310,35 @@ let unclosed opening_name opening_loc closing_name closing_loc =
...
+  type (_, _) t =
+    | Literal : 'ast list -> ('ast, 'ast_desc) t
+    (** A plain array literal/pattern, [[? x; y; z ?]] *)
+    | Opened_literal : open_declaration *
+                       Lexing.position *
+                       Lexing.position *
+                       expression list
+                     -> (expression, expression_desc) t
+    (** An array literal with a local open, [Module.[? x; y; z ?]] (only valid in
+        expressions) *)
```
OlivierNicole, line 322: @antalsz I’m not sure to understand why this is documented as only valid in expressions. Concretely, using the syntax `M.[: 1; 2 :]` works. What am I missing?

OlivierNicole: In addition, there are “unclosed” cases in the grammar `array_exprs` that are not present in `array_patterns`, which seems at odds with the title of this PR. Thanks in advance if you can help me understand this.

goldfirere: @OlivierNicole The comment is about how the pattern grammar doesn't go via this mechanism, though I agree that seems to contradict the title of the PR. It might be worth unifying these approaches as you work with this, or perhaps it's simpler not to use this mechanism to unify the two array syntax treatments -- whatever seems easier.

PR #120, opened 2023-02-09
Handle attributes on lambdas with locally abstract types

ccasin:
Currently we're ignoring attributes on lambda with locally abstract types, like:

```
  let f1 = fun [@inline] (type a) (x : a) -> x
```

The issue is that the attribute ends up on a `Pexp_newtype` field in the parser, which then gets pushed into the `exp_extra` field on the function in the typed tree, and the translation to lambda wasn't collecting attributes from `exp_extra`.

I've fixed it by just collecting attributes from any `Texp_newtype` nodes in the `exp_extra` list before we call `Translattribute.add_function_attributes` in the translation to lambda.  That function is where we check for attributes like `inline` on functions.

I imagine this is an upstream bug as well, and will make a PR there at some point soon.

### DIFF
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -1195,7 +1195,19 @@ and transl_function ~scopes e alloc_mode param cases partial warnings region cur
   let loc = of_location ~scopes e.exp_loc in
   let body = if region then maybe_region body else body in
   let lam = lfunction ~kind ~params ~return ~body ~attr ~loc ~mode ~region in
-  Translattribute.add_function_attributes lam e.exp_loc e.exp_attributes
+  let attrs =
+    (* Collect attributes from the Pexp_newtype node for locally abstract types.
+       Otherwise we'd ignore the attribute in, e.g.;
+           fun [@inline] (type a) x -> ...
+    *)
+    List.fold_left
+      (fun attrs (extra_exp, _, extra_attrs) ->
+         match extra_exp with
+         | Texp_newtype _ -> attrs @ extra_attrs
+         | (Texp_constraint _ | Texp_coerce _ | Texp_poly _) -> attrs)
+      e.exp_attributes e.exp_extra
+  in
+  Translattribute.add_function_attributes lam e.exp_loc attrs
 
 (* Like transl_exp, but used when a new scope was just introduced. *)
 and transl_scoped_exp ~scopes expr =
--- a/lambda/translcore.ml
+++ b/lambda/translcore.ml
@@ -1203,7 +1203,7 @@ and transl_function ~scopes e alloc_mode param cases partial warnings region cur
     List.fold_left
       (fun attrs (extra_exp, _, extra_attrs) ->
          match extra_exp with
-         | Texp_newtype _ -> attrs @ extra_attrs
+         | Texp_newtype _ -> extra_attrs @ attrs
          | (Texp_constraint _ | Texp_coerce _ | Texp_poly _) -> attrs)
       e.exp_attributes e.exp_extra
   in
--- a/testsuite/tests/warnings/w53.ml
+++ b/testsuite/tests/warnings/w53.ml
@@ -378,3 +378,9 @@ end
 
 (* Attributes in attributes shouldn't be tracked for w53 *)
 [@@@foo [@@@deprecated]]
+
+module TestNewtypeAttr = struct
+  (* Check for handling of attributes on Pexp_newtype *)
+  let f1 = fun [@inline] (type a) (x : a) -> x (* accepted *)
+  let f2 = fun [@immediate] (type a) (x : a) -> x (* rejected *)
+end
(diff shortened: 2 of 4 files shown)


testsuite/tests/warnings/w53.ml
```diff
@@ -378,3 +378,9 @@ end
 
 (* Attributes in attributes shouldn't be tracked for w53 *)
 [@@@foo [@@@deprecated]]
+
+module TestNewtypeAttr = struct
+  (* Check for handling of attributes on Pexp_newtype *)
+  let f1 = fun [@inline] (type a) (x : a) -> x (* accepted *)
```
goldfirere, line 384: Is it worthwhile / possible to test that the attribute actually does a thing?

lambda/translcore.ml
```diff
@@ -1195,7 +1195,19 @@ and transl_function ~scopes e alloc_mode param cases partial warnings region cur
...
+  let attrs =
+    (* Collect attributes from the Pexp_newtype node for locally abstract types.
+       Otherwise we'd ignore the attribute in, e.g.;
+           fun [@inline] (type a) x -> ...
+    *)
+    List.fold_left
+      (fun attrs (extra_exp, _, extra_attrs) ->
+         match extra_exp with
+         | Texp_newtype _ -> attrs @ extra_attrs
+         | (Texp_constraint _ | Texp_coerce _ | Texp_poly _) -> attrs)
```
goldfirere, line 1207: Do we want to collect attributes from any of these? e.g. `(fun x -> x : 'a -> 'a) [@inline]`? And could we be missing attributes elsewhere because they end up in the extra? Actually, what's wrong with always using the `extra_attrs`? Even more maximally, why even have the `extra_attrs`? Instead, just combine them with the attrs in the type-checker instead of holding them off to the side. (Maybe because then `Untypeast` can't work.) It just looks to me like you've fixed one bug of a class without addressing the rest.

goldfirere (commented):
I think this fixes the bug you found, but I think that a more general fix will fix more bugs (and be no more complicated).

testsuite/tests/warnings/w53.ml
ccasin: Probably possible, but I think the actual semantic behavior of attributes should probably be tested by attribute-specific tests.  This is a test for warning 53 - the mechanism for giving warnings if the compiler never looks at your attribute.  To figure out how to check that it subsequently does the right thing for each of the ~couple dozen built-in attributes in each of the positions they occur would be a lot of work.

goldfirere: Fair enough.

lambda/translcore.ml
ccasin: I considered this, but decided against it.  Let me see if I can explain why:

- The `exp_extra` type is kind of a grab bag (though I don't know its history).  People could easily add more things to it in the future.  So even if we do decide to take attributes from some or all of the other constructors, we at a minimum want a brittle pattern match here to make people who add constructors in the future think about what those mean when they appear on a function.
- `Texp_newtype` is a pretty special case relative to the others `exp_extra`s, because it's actually part of the function syntax in the parser: `fun (type a) x -> ...`.  If you want to get an `inline` (or other similar attribute) in there without this fix, you have to actually change your function to be `fun (type a) -> fun [inline] x -> ...`, which seems very awkward.
- We certainly could allow you to put `inline`, `specialize`, `noalloc` and friends on type annotations and coercions when those things are on functions, but it feels not in keeping with the current rule that these things go on definition sites.  If we're willing to push `[@inline]` inside a type annotation, why aren't we willing to push it inside an application like `((fun x -> x) 3) [@inline]`?  i think the cleanest user-facing rule is that these must go directly on definitions, and then we need the special case added by this PR because locally abstract types happen to be part of the syntax of function definitions and they are parsed kind of weird.
- I'm a little unsure about the `Pexp_poly` case, as I have limited understanding of the module system, and I have to run, but I can dig in to whether it makes sense later.

I don't feel strongly about the annotation and coercion cases. I do think that given the grab-bag nature of `exp_extra` is makes sense to allow attributes on them to be treated differently, but I don't really know the motivation of the design.

goldfirere: I don't know the original motivation of the design, but it smells similar to something in GHC, so I think I know what's going on. It basically comes down to this: by the time we're in typedtree, we know all the typing information. So source nodes that do nothing but affect typing information are redundant in typedtree -- all that information is already stored in individual expression nodes' types (and pattern nodes, etc.). Accordingly, someone thought it wise (and I might agree) to put these type-information-only nodes off to the side, to make e.g. transl easier to write. (GHC doesn't do this. Instead, the equivalent of transl has to ignore the nodes in a number of places. Not sure what's better.)

The problem is that the type-only nodes might contain attributes, and these attributes should not be off to the side. (To me, this suggests that the nodes themselves shouldn't be off to the side, but it's hard to know all the pros/cons.)

----------------------

Allow me to rephrase your argument above like this: When the user writes `fun [@blah] x y z -> ...`, that `[@blah]` is meant to be an attribute on the `fun` node. If it so happens, though, that `x` is really `(type a)`, then the attribute ends up (uselessly) on a `Pexp_newtype`. Your patch just notices this weirdness and rectifies it.

I agree that `Pexp_poly` is a bit strange... but looking around at its usages, I think it's OK -- that is, I don't think a `Pexp_poly` can ever carry an attribute because of the way the nodes are inserted.

After reading your explanation of what's going on here, I agree with you. This is indeed a special case, deserving of special handling.

goldfirere (approved):

PR #125, opened 2023-02-23
Split out `Clflags.Extension` into a new file, `Language_extension`

antalsz:
This is going to be nicer for integration into ocamlformat and Merlin, I think.  Changes are code movement, renaming references, and updating dependencies.

The build is currently broken for the non-`Makefile.jst` build mechanism, so I haven't run `make depend` yet.


goldfirere (approved):
Looks good. Not merging because of the broken build, which we really must fix, and I'm worried about compounding damage.

riaqn:
FWIW the build is fixed now.

antalsz:
@goldfirere Glad you waited, there were some dependencies that needed updating!  Once CI completes I'll merge, thanks

PR #127, opened 2023-03-01
Cleanup of comprehensions and immutable arrays

antalsz:
This should resolve the last lingering questions about comprehensions and immutable arrays.  It doesn't turn them on by default; I thought that belonged in a separate PR.  @mshinwell, you mentioned you might want more testing; could you take a look at what's here and see if you think there's anything missing?


antalsz:
Ah, and I also stripped out the last uses of `Obj.magic`/`%obj_magic`, as requested by @mshinwell

goldfirere (commented):
* The changes in `lambda` look sensible enough, but I'd prefer @mshinwell or his designee to also look. This should be quick!
* Ditto `middle_end`.
* I've reviewed (and positively like) the changes in `parsing`. Indeed, I may pull those changes out of this PR so that they can land sooner.
* The changes in `stdlib/iarray.ml` surprise me: that's lots of duplication! I personally prefer stating the new functions in terms of the older ones, but I understand @mshinwell may have thought something else was better, and I defer to his judgment and experience.
* Testing, typing, and utils changes are all reviewed and do not need further examination.

goldfirere:
To be clear, my comment above is almost an Approve, but not actually smashing the button without @mshinwell.

mshinwell:
Ack, I'm looking at this.

mshinwell:
I'm a bit uneasy about the `Parray_of_iarray` primitive -- would it be possible to only have the `external` declaration for that in `iarray.ml` (to be used for the copying operation) and avoid exposing this to users?

lpw25:
FWIW It looks to me like if we moved the iarray module out of the stdlib and into a stdlib_extras library -- so that we could use extensions in the stdlib -- then we could replace almost all the uses of `Parray_of_iarray` with comprehensions. The only exceptions I can see are `of_list` and `of_seq`. Implementing `of_seq` as it's own primitive in C would allow us to remove those two efficiently.

One day we'll hopefully support more general comprehensions that allow you to use `for ... in ...` with a sequence to produce an iarray. Another interesting feature might be `[: ... while ... :]` to allow something like:
```ocaml
let of_seq s =
  let first = s () in
  let r = ref first in
  let next () =
    match !r with
    | Nil -> assert false
    | Cons(hd, tl) -> r := tl (); hd
  in
  [: next () while !r <> Nil :]
```

All of this is obviously for another PR though. I think we can live with `Parray_of_iarray` for now.

mshinwell:
@lpw25 do you mean `Parray_to_iarray`, the "freezing" one?  I was referring to the one that treats an immutable array as mutable.

goldfirere:
I had a related thought this morning: I seem to remember from the earlier immutable-arrays patch that the memory allocated for an immutable array is, well, immutable. The new conversion primitives, though, look like they will start treating that memory as an `array`. Will that cause chaos?

mshinwell:
That's the sort of thing I'm concerned about.  Just looking again, if `Iarray.to_array` is the only use for this function, could we use an immutable array -> mutable array duplication primitive instead here?

goldfirere:
Yeah -- having the primitive do the copy feels safer. Except that it might be nice to have e.g. `iter` be implemented for iarray by treating the array as an `array` and then using `Array.iter`. No copying needed here, because we know we're just calling `iter`. So I'm not really sure what's best.

Maybe the cleanest thing is to keep the primitive as written -- no copy -- but carefully label it as non-copying. But it's even more than that: the `array` we get from an `iarray` really had better never be mutated, because it might be immutable memory! (For example, if some function were to create an `iarray`, then convert to an `array`, then do some mutation, then convert back to `iarray`, that would look safe: because the function has just created the thing, we know there are no aliases. But actually it would be disastrous, if the memory were labeled as Immutable.) Still, I think trying to bake too much safety into the primitive will make it too unwieldy. But I'd love a better idea to be surfaced.

mshinwell:
I think I'd prefer to see a copying primitive, then separate implementations for immutable array functions.  The amount of potential code duplication (e.g. for things like `iter`) seems fairly minimal?  This would avoid having complexity in the middle end to ensure that the mutability-breaking operation remains safe.

goldfirere:
> minimal

See https://github.com/ocaml-flambda/ocaml-jst/pull/127/files#diff-016fd769b59a803a3ce3cbafcb7d44f373759d306638297c86dc851e3bd241f3. I would say small or moderate duplication, not quite minimal. It gave me pause, but pauses can be resumed. (That is, the duplication is a red flag, but I understand why it might be the best option in the end.)

@mshinwell: I think you're best positioned to make the final call on how to proceed here, so I nominate you to just make a decision and we'll do what you say. Right now, it looks like: have a copying primitive and keep all the duplicated functions.

Thanks!

goldfirere:
Thanks, @mshinwell. @antalsz, does that give you enough to go on here?

antalsz:
Unless I'm missing something, Mark's request is for the status quo: the only unsafe function the `.mli` exposes is `unsafe_get`.  Everyone's concern about the `%array_of_iarray` primitive seems warranted, but I will observe that `Bytes ` currently uses `%bytes_of_string`; I copied the design from there.  I agree that we could safely use a copying primitive instead, though, for our use cases.

mshinwell:
Sounds like everything's ok then.

antalsz:
Looks like the only conflicts (besides bootstrapping) were from #132 getting pulled out; dropping that commit from here means the rebase went fine!  Will merge after CI completes :partying_face:

PR #131, opened 2023-03-03
Disable module patterns in comprehensions

ncik-roberts:
Module patterns don't quite work in comprehensions. The module bindings aren't entered into the environment (see `f1` below), and it's possible that the typechecking of the `in` list will disable module patterns, anyway (see `f2` below).

For now, @antalsz and I decided to disable these patterns everywhere in comprehensions and to revisit soon.

Before this PR, this was the behavior:
```ocaml
module type S = sig
  type t
  val x : t
end

let t = (module struct
  type t = int
  let x = 3
end : S)

let f1 () =
  [ M.x for (module M : S) in [ t ] ]
(*  ^^^
Error: Unbound module M *)

let f2 () =
  [ M.x
    for (module M :S) in
(*              ^
Error: Modules are not allowed in this pattern. *)
      [ (module struct
          type t = int
          let x = 3
         end : S)
      ]
  ]
``` 

Now, you get the `Error: Modules are not allowed in this pattern.` error in both circumstances, which at least is a straightforward error message.


testsuite/tests/comprehensions/list_comprehensions_modules.ml
```diff
@@ -0,0 +1,43 @@
...
+          let x = 3
+         end : S)
+      ]
+  ];;
+[%%expect {|
+Line 3, characters 16-17:
+3 |     for (module M :S) in
+                    ^
+Error: Modules are not allowed in this pattern.
+|}];;
```
antalsz, line 43: Can you, instead of creating this file, add these tests to all of `{list,array,iarray}_comprehensions_pure.ml`?  They could go nicely at the end of the "`(**** Errors ****)`" section.

(If you'd really prefer it as a separate file, I'd rename it to `comprehensions_module_patterns.ml` and duplicate these tests for both array and immutable array comprehensions, but I think adding them next to the existing tests is cleaner, personally.)

typing/typecore.ml
```diff
@@ -7058,7 +7058,9 @@ and type_comprehension_clauses
 and type_comprehension_clause ~loc ~comprehension_type ~container_type env
   : Extensions.Comprehensions.clause -> _ = function
   | For bindings ->
-      reset_pattern true;
+      (* TODO: fix handling of first-class module patterns so we can pass
+       * "true" here. *)
+      reset_pattern false;
```
antalsz, line 7063: Didn't you also say you had to move this elsewhere?  Or is that only if it's going to be `true`?

antalsz (changes requested):

testsuite/tests/comprehensions/list_comprehensions_modules.ml
ncik-roberts: This is done.

testsuite/tests/comprehensions/array_comprehensions_pure.ml
```diff
@@ -409,3 +409,42 @@ Line 1, characters 23-24:
 Warning 18 [not-principal]: this type-based constructor disambiguation is not principal.
 - : M.t array = [|M.B; M.A; M.A; M.A|]
 |}];;
+
```
antalsz, line 412: You went down a bit too far!  This is the "Test that types are being propagated inwards correctly" section, which is about constructor disambiguation.  You'll want to bring it up above that back into the "Errors" section, and add a leading comment line saying what's in these tests: `(* Module patterns are disallowed (for now) *)`, for instance.  Maybe at the end of the section, maybe right after "No duplicating variables in a for-and clause" (which see for formatting).

antalsz (changes requested):

typing/typecore.ml
ncik-roberts: I don't think we should do this yet -- I tried making the naive change of moving `reset_pattern` to before the call to `type_pat`, but that led to other tests breaking. In other words, it seems non-trivial.

ncik-roberts: Actually I found that this approach was flawed, as, just as the `in` body could disable `allow_modules` mode, so too can it re-enable it. (This means the "unbound module `M`" error could slip through.) Instead, I set the `allow_modules` ref to false just before checking the comprehension pattern.

testsuite/tests/comprehensions/array_comprehensions_pure.ml
ncik-roberts: Actually resolved now :)

antalsz:
LGTM, modulo some formatting concerns; I'll talk to you about those IRL

typing/typecore.ml
```diff
@@ -7137,6 +7137,9 @@ and type_comprehension_iterator
              ~explanation:(Comprehension_in_iterator comprehension_type)
              seq_ty)
       in
+      (* TODO: fix handling of first-class module patterns so we can remove
+       * this line. *)
+      allow_modules := false;
```
goldfirere, line 7142: Why have this here instead of in the call to `reset_pattern` in `type_comprehension_clause`? As it stands, I have to worry about whether some patterns will have modules enabled and others won't... but then the patterns with modules enabled all have type `int`, so I guess it's OK. But much simpler if we just say in `type_comprehension_clause` "no modules here!". But perhaps I've missed an interaction that would make my approach not work...

goldfirere (commented):
I know this has already landed, but I still wanted to take a look.

typing/typecore.ml
ncik-roberts: Calling `reset_patterns false` in [`type_comprehension_clause`](https://github.com/ncik-roberts/ocaml-jst/blob/62ad445d4e9e1fb19dbf97901292d1bdd57faacc/typing/typecore.ml#L7061) doesn't work because, just as typechecking some expressions (read: top-level modules) sets `allow_modules` to `false`, so too does typechecking other expressions (read: let-expressions) sets `allow_modules` to `true`.

In particular, the tests I added in this feature with the `let t = t in t` body would have a worse error message if we did this call in `type_comprehension_clause`.

It's true that I wasn't thinking about the int range case, but that's only because there's no corresponding call to `type_pat`. I didn't confirm this previously, but the error there looks fine:

```
# [ x for (module _ : S) = 1 to 3 ];;
Error: Invalid pattern in comprehension for-range iterator: only variables and _ are allowed.
```

antalsz: Yes, non-variable, non-underscore patterns are syntactically disallowed in `for` loops and, correspondingly, in int ranges in comprehensions

goldfirere: Yes, well, that makes some sense. But it's interesting to note that `let ... and ... and ...` processes all the patterns first, then the RHSs. This allows all the patterns to have one shared `reset_pattern`, with no intervening expressions to process. In contrast, comprehensions processes the clauses one at a time, interleaving patterns with expressions -- meaning the `reset_pattern` approach doesn't work. My hunch is that other problems will arise from the difference in treatment here, in which case it will make sense to refactor this all. (It might be worth reading through `type_let` to see all the stuff that happens between pattern processing and expression processing. But probably not actually worth it -- let's just wait for bugs to crop up.)

@ncik-roberts please Resolve the conversation after reading this. No more action to take here.

PR #133, opened 2023-03-07
Add no-stack-allocation variant of some tests that print lambda

ccasin:
This fixes some tests that were failing with `--enable-stack-allocation=no`.

The tests print lambda code, including the `nlocal` field on functions, which changes depending on allocation configuration.

Alternative solutions include:
- undoing the choice to print nlocal
- adding a flag to not print nlocal
- fixing all the tests that print lambda code to do something smarter

But this seemed easiest.

The CI won't actually test that this works with stack allocation disabled, but I tested it locally.  The immediacy PR will change 32-bit builds to be done without stack allocation, and it will be tested in CI after that.


ccasin:
OK, fixed and ready for review.

An additional option we just discussed in person is to disable 32-bit builds in CI.  I'll make a PR for that, but I guess it's still good to fix test failures like this we happen to notice.

goldfirere (approved):

goldfirere:
There are indeed other design possibilities -- including the drastic choice in #134 of just not doing 32-bit checks at all. But in the interest of unblocking #122, I'll merge this. We can always keep iterating.

PR #134, opened 2023-03-07
Disable 32 bit build in CI

ccasin:
Based on in-person discussion, it seems like there's no reason to be building this in CI.  It has already been disabled in flambda-backend, and we don't use 32-bit builds at all so any failures are only a distraction.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,13 +21,6 @@ jobs:
             use_runtime: d
             ocamlrunparam: "v=0,V=1"
 
-          - name: i386
-            config: CC='cc32' AS='as --32' ASPP='gcc -m32 -c' -host i386-linux PARTIALLD='ld -r -melf_i386'
-            os: ubuntu-20.04
-            ocamlparam: ''
-            boot_config: CC='cc32' AS='as --32' ASPP='gcc -m32 -c' -host i386-linux PARTIALLD='ld -r -melf_i386'
-            boot_cachekey: 32bit
-
     env:
       J: "3"
 


goldfirere:
I'm not a huge fan of this change: it will make upstreaming harder. That is, instead of the author of a patch learning in a timely fashion that their change breaks 32-bit, the poor beleaguered upstreamer will encounter the problem, probably while they're porting a small pile of patches all at once. That's much harder to debug! Yes, keeping this CI slows us down, but I think the net slowdown of not checking will prove to be larger (even though the slowdown comes later, and therefore costs less).

ccasin:
> I'm not a huge fan of this change: it will make upstreaming harder.

A couple quick points:
- The i386 backend is dead upstream and won't be coming back.  If we do want to test a 32-bit build, we should test the bytecode one.
- The decision in flambda-backend was to stop testing the 32-bit compiler, and the plan remains to eliminate this repo and move to that one, so I suppose the real question is whether to add back a 32-bit bytecode compiler there.
- I believe we've stopped testing even our public release code on 32-bits.  I'll separately send you a link to the relevant internal slack channel.

So while I can see the case for changing flambda-backend to test a 32-bit bytecode compiler, I think until the decision is revisited there it doesn't make much sense to bother to do it here.

goldfirere:
I agree with @ccasin above -- but with the difference that I do think we should re-enable this in flambda-backend (if/when we all move into that repo). Right now, there is no upstream plan for flambda-backend, so having the tests off makes good sense. But if flambda-backend grows to include being host to patches that we do plan to upstream, I think it's best to test on 32-bit bytecode.

I don't feel strongly here because I see pluses and minuses in both directions; just wanted to state my case.

ccasin:
Subsumed by #155

PR #135, opened 2023-03-07
Remove `Lev_module_definition` from lambda

ncik-roberts:
@mshinwell says he originally added the `Lev_module_definition` construct in an earlier aborted attempt for GDB support. Removing it altogether makes some things simpler for me in my patch to make module pattern unpacking more sensible. In this PR, I remove it. This has no effect on bytecode and just affects the intermediate lambda AST.

### DIFF
--- a/.depend
+++ b/.depend
@@ -306,6 +306,7 @@ parsing/ast_mapper.cmo : \
     parsing/longident.cmi \
     parsing/location.cmi \
     utils/load_path.cmi \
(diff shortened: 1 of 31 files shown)


mshinwell (approved):

mshinwell:
Yeah I don't think we need this now, as the extra scoping information in `Debuginfo.t` that was added for memtrace should suffice.

PR #138, opened 2023-03-08
Basic `unique` Extension

riaqn:
This PR is a simplified version of https://github.com/ocaml-flambda/ocaml-jst/pull/116 where overwrite (`{overwrite_ x with ...}`) is removed. The `overwrite` feature requires changes in the middle-end and back-end which is easily unsafe if not done right. This PR doesn't include `overwrite` and should be easier to review and merge. Moreover, most of the race-freedom guarantee we aim to provide doesn't rely on overwriting anyway. 

This PR still touches the middle and back-end because
- Most are the renaming from `alloc_mode` to `locality_mode`. Now `alloc_mode` contains two axes: `locality` and `uniqueness`.
- The `uniqueness` axis is passed through, because they might be needed even in middle and back-end. 
Currently we only use it to disable constant-lifting for `unique` values in multiple stages in the pipeline.


typing/typecore.ml
```diff
@@ -4106,27 +4268,56 @@ and type_expect_
...
+        raise (Typetexp.Error (loc, Env.empty, Unsupported_extension Unique));
+      let mode = if mode_cross env ty_expected then mode_max
+      else begin
+        let mode = mode_once expected_mode in
+        submode ~loc ~env ~reason:Other mode.mode expected_mode;
+        mode
+      end
+      in
+      let exp = type_expect ?in_function ~recarg env mode sbody ty_expected_explained in
+      { exp with exp_loc = loc }
```
riaqn, line 4305: I'm unsure about this - when we write `once_ exp` do we mean that `exp` needs to be strictly `once`? Or is `many` also ok?

lambda/translcore.ml
```diff
@@ -220,7 +220,7 @@ let rec push_defaults loc bindings use_lhs arg_mode cases partial warnings =
             ({exp with exp_type = pat.pat_type; exp_env = env; exp_desc =
               Texp_ident
                 (Path.Pident param, mknoloc (Longident.Lident name),
-                 desc, Id_value)},
+                 desc, Id_value, (Mode.Uniqueness.shared, Mode.Linearity.many))},
```
lpw25, line 223: I think this mode will never be consumed, but using `mode` here would be more correct than this dummy value.

typing/typecore.ml
lpw25: `once_ exp` does not make any requirement on `exp` -- it expects it to produce a `once` value but that's the weakest mode so doesn't change anything.

lpw25 (commented):
I read through most of the changes to the back and middle-end, but I actually think it would be better to take those out since they can't really be tested in this PR. Better to have a PR that obviously doesn't change compilation, and then later change compilation when we have operations that require those changes.

typing/typecore.ml
```diff
@@ -5290,9 +5494,10 @@ and type_expect_
                    ; _ }
                   , _)}]) ->
         check_probe_name name name_loc env;
-        let env = Env.add_lock Alloc_mode.global env in
+        let env = Env.add_locality_lock Mode.Locality.global env in
+        (* TODO: do we need linearity lock here? *)
```
riaqn: Please check: do we need linearity lock here?

typing/typecore.ml
```diff
@@ -4183,18 +4388,19 @@ and type_expect_
...
              parent's region having a return mode of local. *)
-          submode ~loc ~env ~reason:Other Value_mode.local mode';
+          submode ~loc ~env ~reason:Other {mode'.mode with locality = Mode.Regionality.local} mode';
           let new_env = Env.add_exclave_lock env in
           let exp =
             type_expect ?in_function ~recarg new_env mode' sbody ty_expected_explained
           in
-          submode ~loc ~env ~reason:Other Value_mode.regional expected_mode;
+          submode ~loc ~env ~reason:Other {expected_mode.mode
+            with locality = Mode.Regionality.regional} expected_mode;
```
riaqn, line 4403: this submode (and the one above) looks ugly right now. They will be better with `min_with_locality` in https://github.com/ocaml-flambda/ocaml-jst/pull/160.

riaqn:
This PR is now rebased, and stripped off most middle-end and back-end changes.

typing/typecore.ml
```diff
@@ -2605,7 +2719,7 @@ let partial_pred ~lev ~splitting_mode ~allow_modules ?(explode=0)
       } in
   try
     reset_pattern allow_modules;
-    let alloc_mode = simple_pat_mode Value_mode.global in
+    let alloc_mode = simple_pat_mode (Mode.Value.legacy) in
```
riaqn, line 2722: not sure if this should be `legacy` or `min`

typing/typecore.ml
```diff
@@ -1287,7 +1403,7 @@ let type_for_loop_like_index ~error ~loc ~env ~param ~any ~var =
   | Ppat_any -> any (Ident.create_local "_for")
   | Ppat_var name ->
       var ~name
-          ~pv_mode:Value_mode.global
+          ~pv_mode:Mode.Value.legacy
```
riaqn, line 1406: I believe this is `min` because of integer mode crossing.

lpw25 (commented):
I've reviewed everything outside of the testsuite and typing directories. I've also reviewed `types.mli` and `mode.mli`. Since you're away, I've addressed my various comments in a commit on top of your branch which I will push shortly.

testsuite/tests/typing-unique/unique_mod_class.ml
```diff
@@ -0,0 +1,150 @@
...
+  let module Bar = struct
+    let y = unique_ x
+  end
+  in
+  ()
+[%%expect{|
+Line 4, characters 12-21:
+4 |     let y = unique_ x
+                ^^^^^^^^^
+Error: This value is used as unique but it is a value that is top-level
```
lpw25, line 57: This error message is confusing. I cannot decipher from this what the actual restriction is.

lpw25 (commented):
I've reviewed the latest changes and also the changes in the testsuite. The changes in the typing directory remain unreviewed.

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -0,0 +1,465 @@
...
+  | x :: xs as gs -> (gs, xs)
+[%%expect{|
+Line 3, characters 22-24:
+3 |   | x :: xs as gs -> (gs, xs)
+                          ^^
+Error: This is used uniquely here so cannot be used twice.  Another use is
+Line 3, characters 26-28:
+3 |   | x :: xs as gs -> (gs, xs)
+                              ^^
+  The latter is the position-1 element of the constructor "::" in the value which is the former
```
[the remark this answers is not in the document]
riaqn: Yeah I think for simple cases (e.g. direct parent-child) it's better to keep it simple; but on the other hand, I'm afraid that in the complex case of multi-level relation the user would want to know the exact relation between the parent and the child.

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -0,0 +1,465 @@
...
+[%%expect{|
+Line 2, characters 12-13:
+2 |   match (a, b) with
+                ^
+Error: This is used uniquely here so cannot be used twice. It is in a tuple matched against a variable. Another use is
+Line 3, characters 28-30:
+3 |   | (true, b') -> unique_id b'
+                                ^^
+
+|}]
```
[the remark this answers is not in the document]
riaqn: This example is an error because of `as _t` causes the usage of both `a` and `b`, which clashes with the unique use of `b` in the first branch. While intuitively this still shouldn't be an error, because the construction of `_t` (and hence the usage of `a` and `b`) happens only in the second branch which shouldn't interfere with the first branch, the current implementation views this differently. Upon seeing this `as _t`, both `a` and `b` will be used,  a new tree root is generated corresponding to `_t`, and further patterns `(false, b')` will be matched against this new tree root. It is therefore difficult to separate the usage caused by `false` and the usage caused by `_t`.
A better approach might be somehow to link `_t` to the paths of `a` and `b`. 

For now I will add a comment to ask for future improvement.

testsuite/tests/typing-unique/unique_mod_class.ml
riaqn: How about:
> Error: This value is used as unique but it is a value that is outside of the current module or class.

typing/env.mli
```diff
@@ -214,9 +224,16 @@ val lookup_error: Location.t -> t -> lookup_error -> 'a
...
+    the returned uniqueness mode is too high because of some linearity_lock during
+    lookup, and fail to satisfy expected_mode in the caller.
+
+    TODO: A better approach is passing down the expected mode to this
+    function as argument, so that sub-moding error is triggered at the place where
+    error hints are immediately available. *)
 val lookup_value:
   ?use:bool -> loc:Location.t -> Longident.t -> t ->
-  Path.t * value_description * Types.value_mode
+  Path.t * value_description * Mode.Value.t * shared_context list
```
lpw25, line 236: Maybe just return the most recent `shared_context`? Presumably only one of them is actually used to give a hint.

typing/typetexp.ml
```diff
@@ -455,7 +475,15 @@ and transl_type_aux env policy mode styp =
...
           let arg_cty = transl_type env policy arg_mode arg in
-          let acc_mode = Alloc_mode.join_const acc_mode arg_mode in
+          let acc_mode =
+            Alloc.Const.join
+              (Alloc.Const.uncurried_ret_mode_from_arg arg_mode)
+              (Alloc.Const.uncurried_ret_mode_from_alloc acc_mode)
+          in
+          let acc_mode =
+            Alloc.Const.join acc_mode
+              (Alloc.Const.min_with_uniqueness Uniqueness.Const.Shared)
```
lpw25, line 485: Is this needed? Isn't `uncurried_ret_mode_from_alloc` always `Shared` so the join should also already be `Shared`?

lpw25 (commented):
I've now reviewed everything except `typecore` and `uniqueness_analysis`

testsuite/tests/typing-unique/unique_mod_class.ml
riaqn: I like that. I'm not sure how to implement it though - do we construct a `exception Typecore.Error`? I would also add a constructor to `Typecore.error` for this new type of error. I'm fine with that - just want to make sure that this is ok.

typing/mode.mli
```diff
@@ -0,0 +1,350 @@
...
+type ('a, 'b, 'c) modes = { locality : 'a; uniqueness : 'b; linearity : 'c }
+
+module Alloc : sig
+  module Const : sig
+    type t = (Locality.Const.t, Uniqueness.Const.t, Linearity.Const.t) modes
+
+    val legacy : t
+    val join : t -> t -> t
+    val uncurried_ret_mode_from_arg : t -> t
+    val uncurried_ret_mode_from_alloc : t -> t
```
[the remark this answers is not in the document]
riaqn: Yes they are great names - succinct and accurate. I like it.

typing/oprint.ml
```diff
@@ -274,13 +274,80 @@ let pr_var = Pprintast.tyvar
...
+  { oam_locality = join_locality m1.oam_locality m2.oam_locality;
+    oam_uniqueness = join_uniqueness m1.oam_uniqueness m2.oam_uniqueness;
+    oam_linearity = join_linearity m1.oam_linearity m2.oam_linearity }
+
+let default_mode =
+  { oam_locality = Olm_global;
+    oam_uniqueness = Oum_shared;
+    oam_linearity = Olinm_many; }
+
+let uncurried_ret_mode_from_arg arg_mode =
```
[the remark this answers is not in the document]
lpw25: In another comment I suggested `closure_mode` for this one and `partial_app_mode` for the other.

typing/oprint.ml
```diff
@@ -303,6 +370,45 @@ let rec print_out_type_0 mode ppf =
...
+    if is_unique then begin
+      pp_print_string ppf "unique_";
+      pp_print_space ppf () end;
+    if is_once then begin
+      pp_print_string ppf "once_";
+      pp_print_space ppf () end;
+    print_out_type_2 mode ppf ty end
+  else
+    (* otherwise we would rather print everything in attributes
+       even if extensions are enabled *)
```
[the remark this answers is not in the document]
lpw25: Ok, maybe this is fine then.

testsuite/tests/typing-unique/unique_mod_class.ml
lpw25: Yeah, it's ok. There are a few other examples of files raising error messages from other files.

typing/typetexp.ml
riaqn: Both `close_over` and `partial_apply` give `unique` because I think they are used as lower bound, and we don't want to constrain the uniqueness axis of functions. When interpreting type signatures, however, we want functions always to be `shared`. That means `Mode.close_over` and `Mode.Const.close_over` would have different behaviours on the uniqueness axis. I feel I should avoid that, so the two have the same behaviours, but the caller will do some extra tweaking.

testsuite/tests/typing-unique/unique_mod_class.ml
riaqn: Ahh I just realized that there might be some circular dependency issues between `typecore.ml` and `uniqueness_analysis.ml` - I think it might just be easier to construct the message from scratch.

testsuite/tests/typing-unique/unique_analysis.ml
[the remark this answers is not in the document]
riaqn: We discussed last time that I could try to print the path in ocaml syntax:
> Error: Cannot use `xs` here as it has already been used uniquely as part of `gs`, where `gs = (xs, _)`

But in retrospect this would add some extra complexity (for example, the projection need to include the # of fields in the tuple) so I think it's not worth it. I will just do the simple stuff.

typing/typecore.ml
```diff
@@ -4244,18 +4454,23 @@ and type_expect_
...
           raise (Error (loc, env, Exclave_in_nontail_position))
-        | RTail (mode, _) ->
+        | RTail (regionality, _) ->
           (* mode' is RNontail, because currently our language cannot construct
              region in the tail of another region.*)
-          let mode' = mode_default mode in
+          let mode' =
+            mode_default
+              (Value.with_locality regionality expected_mode.mode)
+          in
```
lpw25, line 4463: Maybe add a `mode_exclave` function that does the above.

typing/typecore.ml
```diff
@@ -4244,18 +4454,23 @@ and type_expect_
...
           (* The middle-end relies on all functions which allocate into their
              parent's region having a return mode of local. *)
-          submode ~loc ~env ~reason:Other Value_mode.local mode';
+          submode ~loc ~env ~reason:Other
+            (Value.min_with_locality Regionality.local) mode';
```
lpw25, line 4467: With the new solver I think we should have submodules for e.g. `Value.Left` and `Value.Right`. Then we can have things like `Value.Left.local` for `Value.min_with_locality Regionality.local`.

typing/typecore.ml
```diff
@@ -5165,8 +5386,9 @@ and type_expect_
       let to_unify = Predef.type_lazy_t ty in
       with_explanation (fun () ->
         unify_exp_types loc env to_unify (generic_instance ty_expected));
-      let env = Env.add_lock Alloc_mode.global env in
-      let arg = type_expect env mode_lazy e (mk_expected ty) in
+      let env = Env.add_locality_lock Locality.global env in
+      (* the thunk will be called only once, so no linearity lock needed *)
```
lpw25, line 5390: This is true now, but not in OCaml 5 due to potential races. For now let's type it like it will run many times since it will backwards compatible to be more accurate later.

typing/typecore.ml
```diff
@@ -5433,9 +5660,9 @@ and type_expect_
                    ; _ }
                   , _)}]) ->
         check_probe_name name name_loc env;
-        let env = Env.add_lock Alloc_mode.global env in
+        let env = Env.add_locality_lock Locality.global env in
```
lpw25, line 5663: Probably better to mark this as running `many` too.

typing/typecore.ml
```diff
@@ -7611,6 +7883,43 @@ let escaping_hint failure_reason submode_reason
...
+          because it was defined outside of the let-op.@]"]
+  | Some Env.Class ->
+    [Location.msg
+        "@[Hint: This identifier cannot be used uniquely,@ \
+          because it is defined in a class.@]"]
+  | Some Env.Closure ->
+    [Location.msg
+        "@[Hint: This identifier was defined outside of the current closure.@ \
+          Either this closure has to be once, or the identifier can be used only@ \
+          as shared @]"]
```
lpw25, line 7913: Missing `.` at the end of the message

typing/typecore.ml
```diff
@@ -8076,14 +8393,25 @@ let report_error ~loc env = function
...
+  | Uncurried_function_escapes e ->
       Location.errorf ~loc
-        "This function or one of its parameters escape their region@ \
+        "This function or one of its parameters %s @ \
          when it is partially applied"
+        (match e with
+        | `Locality -> "escape their region"
+        | `Uniqueness -> assert false
+        | `Linearity -> "violates once constraint"
+        )
```
lpw25, line 8414: The would be easier to read if the string was let bound. "violates once constraint" is a bit hard to read. How about "will be used 'many' when then were expected to be used 'once'"?

lpw25 (commented):
Reviewed `typecore.ml`

typing/typecore.ml
riaqn: Yeah that sounds like a good idea. But I think `Value.local` is sufficient, because "being left" should be encoded in the types already.

typing/typecore.ml
```diff
@@ -4198,42 +4370,80 @@ and type_expect_
...
-      ({ pexp_desc = Pexp_extension({txt = ("ocaml.local" | "local" | "extension.local" as txt)}, PStr []) },
+      ({ pexp_desc = Pexp_extension({
+            txt = ("ocaml.unique" | "unique" | "extension.unique" as txt)}, PStr []) },
+       [Nolabel, sbody]) ->
+      if txt = "extension.unique" && not (Language_extension.is_enabled Unique) then
+        raise (Typetexp.Error (loc, Env.empty, Unsupported_extension Unique));
+      let mode = if mode_cross env ty_expected then mode_max
+      else begin
+        let mode = mode_unique expected_mode in
+        submode ~loc ~env ~reason:Other mode.mode expected_mode;
```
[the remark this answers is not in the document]
riaqn: yeah it's not needed - I just wrote it like that so it looks similar to the other cases. But I'm also fine with removing it.

typing/typecore.ml
```diff
@@ -4018,7 +4172,22 @@ and type_expect_
...
   | None      -> match sexp.pexp_desc with
   | Pexp_ident lid ->
-      let path, mode, desc, kind = type_ident env ~recarg lid in
+      let path, mode, shared_reason, desc, kind = type_ident env ~recarg lid in
+      let mode =
+        match path with
+        | Path.Pident _ ->
+            let uniqueness, _ =
+              Uniqueness.newvar_above
+                (Value.uniqueness mode)
```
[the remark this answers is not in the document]
riaqn: That makes sense but I think there is an issue: `mode` retreived from the environment is a "left" value, but later we might want to say `shared <= mode` if we want to enforce it. Indeed, this will be not even typable once the new solver kicks in.

typing/typecore.ml
```diff
@@ -3112,12 +3225,14 @@ let type_omitted_parameters expected_mode env ty_ret mode_ret args =
...
-                    Value_mode.regional_to_local_alloc marg)
+                    Value.regional_to_local_alloc marg)
                  open_args
              in
              let closed_args = new_closed_args @ closed_args in
              let open_args = [] in
-             let mode_closure = Alloc_mode.join (mode_fun :: closed_args) in
+             let m0 = List.map Alloc.close_over closed_args in
+             let m1 = Alloc.partial_apply mode_fun in
+             let mode_closure, _ = Alloc.newvar_above (Alloc.join (m1 :: m0)) in
```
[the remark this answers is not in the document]
riaqn: I can't remember why I did this, but it might be that `Alloc.join` sometimes gives the _least_ upper bound, instead of just upper bound, and therefore sometimes it's not pushed to `local` when it should.

In the new solver, `join` gives a "left" mode, while "register_allocation_mode" would require a "right" mode (because you want to say `local <= m`. So this `newvar_above` is needed. There are several occurrences in the new solver PR where I have to make new vars to type check. Unfortunately I didn't backport each of them here, leaving potential BUGs.

typing/env.ml
```diff
@@ -2901,32 +2916,55 @@ let lookup_ident_module (type a) (load : a load) ~errors ~use ~loc s env =
...
+          );
+          (* outside unique values can be accessed inside a loop but only at shared *)
+          let min_uniq =
+            Mode.Uniqueness.join
+              [ Mode.Value.uniqueness vmode;
+                Mode.Linearity.to_dual mode ]
+          in
+          let vmode = Mode.Value.with_uniqueness min_uniq vmode in
+          vmode, Some shared_context
+        end
```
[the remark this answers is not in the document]
riaqn: I like the generality of this, but also feel it would be nicer if implemented with the new solver - can I do it there instead? (considering that the code is already correct, and the new solver PR is not far from merging)

lpw25 (commented):
Some more comments

typing/env.ml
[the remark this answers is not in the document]
lpw25: Yeah, that's fine. You should still make the change to have a shared `Closure_lock` in this branch though.

riaqn:
(reminder to myself that `shared_context` should be removed from `expected_mode`)

typing/typecore.ml
lpw25: I think `min` would be better here.

typing/typecore.ml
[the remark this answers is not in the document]
lpw25: Actually, I think this whole part is needlessly confusing. Fundamentally, we place two kinds of restrictions on modes from the uniqueness analysis:
1. We constrain the variable to be used `shared` -- that means we need a "right" uniqueness mode.
2. We constraint the variable to be defined `many` -- that means we need a "left" linearity mode.
We already have such modes available without creating any variables: we can use the uniqueness of the current expected mode, and the linearity that comes from the environment.

In the case that the extension is disabled we can just constrain those two modes to be `shared` and `many` respectively. This also avoids the need to distinguish the `Path.Pident` case since we won't be creating any extra variables.

typing/typecore.ml
```diff
@@ -7910,11 +7906,19 @@ let sharedness_hint _fail_reason submode_reason context =
...
-          as shared @]"]
+          as shared.@]"]
   | Some Env.Module ->
     [Location.msg
         "@[Hint: This identifier cannot be used uniquely,@ \
           because it is defined in a module.@]"]
+  | Some Env.Probe ->
+    [Location.msg
+        "@[Hint: This identifier cannot be used uniquely,@ \
+          because it is defined outside of the class.@]"]
```
lpw25, line 7917: s/class/probe/

typing/typecore.ml
```diff
@@ -8404,14 +8408,15 @@ let report_error ~loc env = function
...
-        "This function or one of its parameters %s @ \
-         when it is partially applied"
-        (match e with
+      let reason  =
+        match e with
         | `Locality -> "escape their region"
         | `Uniqueness -> assert false
-        | `Linearity -> "violates once constraint"
-        )
+        | `Linearity -> "will be at once but expected to be many"
```
lpw25, line 8415: "is once but will be used many"

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+  (* lift par and seq to trees *)
+
+  let rec par t0 t1 =
+    {
+      usage = Usage.par t0.usage t1.usage;
+      children =
+        Projection.Map.merge
+          (fun _proj c0 c1 ->
+            let c0 = Option.value c0 ~default:t0 in
+            let c1 = Option.value c1 ~default:t1 in
```
lpw25, line 439: There is some opportunity to short-circuit some cases here. For instance, `t1.usage` < `t0.usage` and `c1` is `None` then we can just return `c0`.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+        let t1 = Option.value t1 ~default:UsageTree.empty in
+        Some (UsageTree.par t0 t1))
+      t0 t1
+
+  let seq : t -> t -> t =
+   fun t0 t1 ->
+    Root_id.Map.merge
+      (fun _rootid t0 t1 ->
+        let t0 = Option.value t0 ~default:UsageTree.empty in
+        let t1 = Option.value t1 ~default:UsageTree.empty in
```
lpw25, line 563: As with `par`, `seq` with an `empty` tree is a potentially expensive no-op.

lpw25 (commented):
More comments

typing/typecore.ml
```diff
@@ -287,13 +291,14 @@ type region_position =
      together with the mode of that region,
      and whether it is also the tail of a function
      (for tail call escape detection) *)
-  | RTail of Value_mode.t * function_position
+  | RTail of Regionality.t * function_position
 
 type expected_mode =
   { position : region_position;
     escaping_context : Env.escaping_context option;
+    shared_context : Env.shared_context option;
```
riaqn, line 299: This field is not really needed.

typing/typecore.ml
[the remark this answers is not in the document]
riaqn: Yeah that makes perfect sense.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+        | Construct_field of string * int
+        | Variant_field of label
+        (* TODO: it would be nicer to remove this memory_address thing for the
+           following reason: Memory_address having Memory_address as child is
+           non-sense and yet representable. Another reason: awkward in error
+           message printing. The root cause of both reasons is that
+           foo.Memory_address is not a child of foo, but foo itself.
+
+           Instead we should represent Memory_address as a property on the value
+           itself (instead of as its child). *)
```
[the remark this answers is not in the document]
riaqn: There is still the problem that `memory_address` is invisible from user's perspective - `x` and `x.mem_addr` print the same - and this special treatment of `mem_addr` just kinda bugs me. Maybe that's just me. Also I think in terms of implementation it's not that different.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+open Types
+open Typedtree
+module Uniqueness = Mode.Uniqueness
+module Linearity = Mode.Linearity
+
+module Occurrence = struct
+  (* The occurrence of a potentially unique ident in the expression. *)
+  type reason =
+    | DirectUse (* it is used directly *)
+    | MatchTupleWithVar of Location.t
```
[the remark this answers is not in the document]
riaqn: I feel the case of `MatchTupleWithVar` could be hard to decipher? For example:
```
match x,y with
|....
| z -> ...
| ....
```
In this case, I feel pointing to `z` as the culprit might be useful. (Currently the location of `z` is not printed - but I can make it so if we decide it's good to do so).

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+     - shared
+     - shared or unique
+     - error
+
+     error is represented as exception which is just easier.
+  *)
+  type t =
+    | Unused
+    (* no good reason to factor out the two, except just to separating the code
+    *)
```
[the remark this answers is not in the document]
riaqn: I will merge everything - I don't think `MaybeShared` and `MaybeBorrow` are worth their module, because they are closed under `seq`. For example, `seq MaybeShared MaybeShared` = `Shared`.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+            let uniqs = List.map (fun ((uniq, _), _) -> uniq) l1 in
+            (* if any of l1 is unique, then all of l0 must be borrowed *)
+            let uniq = Mode.Uniqueness.meet uniqs in
+            List.iter
+              (fun (barrier, _) ->
+                match !barrier with
+                | Some _ -> assert false
+                | None -> barrier := Some uniq)
+              l0;
+            SharedUnique m1)
```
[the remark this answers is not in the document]
riaqn: (I think the example you means is `seq (seq (MaybeShared ...) (MaybeUnique ...)) (MaybeUnique...)`)

That sounds reasonable, but slightly hard to implement in the current framework - the inner `seq` already returns `MaybeUnique` - do we require it to carry a `unique_barrier` as well? I think that might complicate the code but doesn't bring much.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+               the type checking of the whole file, l1 will correctly tells
+               whether it needs to be unique, and by extension whether l0 can be
+               shared. *)
+            let uniqs = List.map (fun ((uniq, _), _) -> uniq) l1 in
+            (* if any of l1 is unique, then all of l0 must be borrowed *)
+            let uniq = Mode.Uniqueness.meet uniqs in
+            List.iter
+              (fun (barrier, _) ->
+                match !barrier with
+                | Some _ -> assert false
```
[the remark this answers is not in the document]
riaqn: Yeah I believe each barrier (associate to each field projection) is written to only once. The content is strictly the "maybe unique" usage immediately following the field projection. I think it's a nice property and I should make it explicit by `assert false`. 

Also, I'm not familiar with the code-motion, but I wonder if it's flexible to express "code-motion but not beyond this point" instead of "no code-motion".

> Could we just have a list of modes on the barrier and treat them as a meet? It might be more efficient that calling actual meet anyway, especially if we included some short-circuiting for the case of constant modes.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+        match (m0, m1) with
+        | Borrowed _, Shared _ -> SharedUnique m1
+        | Borrowed _, MaybeUnique _ -> SharedUnique m1
+        | MaybeShared _, Shared _ -> SharedUnique m1
+        | MaybeShared l0, MaybeUnique l1 ->
+            (* Four cases:
+               B;S = S
+               B;U = U
+               S;S = S
+               S;U /=
```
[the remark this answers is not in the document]
riaqn: Do you mean I should spell out the full name of the modes? I.e. `Borrowed` instead of `B`?

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+      with SharedUnique.MultiUse { here; there; error } ->
+        (* t' is ancestor *)
+        let t', path =
+          match (projs0, projs1) with
+          | [], _ -> (t1.usage, projs1)
+          | _, [] -> (t0.usage, projs0)
+          | _, _ -> assert false
+        in
+        let there_is_of_here =
+          if (Usage.extract_occurrence t').loc = here.loc then Descendant
```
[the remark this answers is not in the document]
riaqn: I was on the extreme side of modularization. There are two `MultiUse` exception, one in `Usage` and the other in `UsageTree`. The latter is the former lifted to the tree structure and thus contains the relative path between the two conflicting occurrences. The code you are refering to is deciding the ancestor-descendant relation between the two.

This code is now removed, since we no longer print the path between the two occurrences.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+            let c0 = Option.value c0 ~default:t0 in
+            let c1 = Option.value c1 ~default:t1 in
+            Some (par c0 c1))
+          t0.children t1.children;
+    }
+
+  (* projs0 is the list of projections for t0.
+     It is from the ancestor that t0.usage actually comes from,
+     to t0.*)
+  let rec seq t0 projs0 t1 projs1 =
```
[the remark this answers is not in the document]
riaqn: In fact, now that we no longer print the relative paths between two occurrences, the whole business about `projs0/projs1` can just be removed - or should I keep it in case it's needed in the future?

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+        | `Linearity -> "defined as once"
+      in
+      let there_reason =
+        match there.reason with
+        | DirectUse -> Format.dprintf ""
+        | MatchTupleWithVar _loc' ->
+            Format.dprintf "which is in a tuple matched against a variable"
+      in
+      let relation =
+        if UsageTree.Path.real_length path = 0 then Format.dprintf ""
```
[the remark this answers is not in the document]
riaqn: Indeed - I no longer print who is a part of who, but simply say that they overlap in some way.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+      (* we are constructing a closure here, and therefore any borrowing of free
+         variables in the closure is in fact using shared. *)
+      let uf' =
+        UF.map
+          (function
+            | BorrowedShared (MaybeShared _ as u) ->
+                (* only implicit borrowing lifted. *)
+                SharedUnique
+                  (SharedUnique.Shared (BorrowedShared.extract_occurrence u))
+            | _ -> Unused)
```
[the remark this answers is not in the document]
riaqn: The root node has the highest mode in the tree (think `Unused`), and `par Unused x` gives `x` - I can't see the issue here?

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+    | proj :: path ->
+        {
+          usage = Unused;
+          children = Projection.Map.singleton proj (singleton path leaf);
+        }
+
+  let map f t =
+    let rec loop acc t =
+      (* to ensure the tree invariant: children >= parent *)
+      let usage = Usage.par acc (f t.usage) in
```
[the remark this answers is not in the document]
riaqn: Hmm how about we use set instead of list for `maybe_unique`? That should de-duplicate automatically?

typing/typecore.ml
```diff
@@ -4198,42 +4370,80 @@ and type_expect_
...
+      [Nolabel, sbody]) ->
+      if txt = "extension.once" && not (Language_extension.is_enabled Unique) then
+        raise (Typetexp.Error (loc, Env.empty, Unsupported_extension Unique));
+      let mode = if mode_cross env ty_expected then mode_max
+      else begin
+        let mode = mode_once expected_mode in
+        submode ~loc ~env ~reason:Other mode.mode expected_mode;
+        mode
+      end
+      in
```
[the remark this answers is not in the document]
riaqn: This means `once_  42` would trigger error if the expected_mode is `many` - is that what we want? Currently we have `local_ 42` type check even if the expected_mode is `global`.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+    | Unused, m | m, Unused -> m
+    | BorrowedShared m0, BorrowedShared m1 ->
+        BorrowedShared (BorrowedShared.par m0 m1)
+    | BorrowedShared _, m | m, BorrowedShared _ ->
+        m (* m must be sharedunique *)
+    | SharedUnique m0, SharedUnique m1 -> SharedUnique (SharedUnique.par m0 m1)
+
+  (* sequential composition *)
+  let seq m0 m1 =
+    match (m0, m1) with
```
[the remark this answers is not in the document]
riaqn: It was inserted by ocamlformat, but yes it looks wierd.

typing/env.ml
```diff
@@ -2921,32 +2935,55 @@ let lock_mode ~errors ~loc env id vmode locks =
...
+              (Mode.Regionality.global)
           with
           | Ok () -> (vmode, reason)
           | Error _ ->
               may_lookup_error errors loc env
-                (Local_value_used_in_closure (id, escaping_context))
+                (Local_value_escaping (id, escaping_context))
+        end
+      | Share_lock shared_context -> begin
+          ( match
```
lpw25, line 2950: Weird space between `match` and `(`. Also, you don't need both the `begin ... end` and the parentheses.

lpw25 (commented):
Reviewed changes since last review

typing/uniqueness_analysis.ml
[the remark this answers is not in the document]
riaqn: Yeah but right after the `UF.map`, the result is `par`-ed with the original `uf`. So that seems fine?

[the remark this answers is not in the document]
lpw25: It might be correct but I don't think it's fine. The correctness is not at all obvious. I really think it would be better to have a simple `UF.map` function that only works on monotonic functions, then to change this function to be monotonic by returning it's input in the `_` case, and then getting rid of the call to `par` afterwards as it won't be needed.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+  let rec seq t0 projs0 t1 projs1 =
+    let usage =
+      try Usage.seq t0.usage t1.usage
+      with SharedUnique.MultiUse { here; there; error } ->
+        (* t' is ancestor *)
+        let t', path =
+          match (projs0, projs1) with
+          | [], _ -> (t1.usage, projs1)
+          | _, [] -> (t0.usage, projs0)
+          | _, _ -> assert false
```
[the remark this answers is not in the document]
riaqn: This comment is resolved because I'm removing the whole logic about projs.

typing/uniqueness_analysis.ml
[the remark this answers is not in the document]
riaqn: This is resolved because `UF.map` now requires monotonic functions.

typing/typecore.ml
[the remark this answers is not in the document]
riaqn: I fixed the code and it also has mode crossing. Please check if it looks good.

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -0,0 +1,465 @@
...
+  let t = update r in
+  t
+[%%expect{|
+Line 3, characters 18-19:
+3 |   let _s = update r in
+                      ^
+Error: This is used uniquely here so cannot be used twice.  Another use is
+Line 4, characters 17-18:
+4 |   let t = update r in
+                     ^
```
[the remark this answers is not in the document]
riaqn: The comment is resolved because there is a newer comment on the same issue.

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -86,58 +86,64 @@ Line 4, characters 17-18:
...
+                                  ^^
 Error: This is used uniquely here so cannot be used twice.  Another use is
-Line 3, characters 26-28:
-3 |   | x :: xs as gs -> (gs, xs)
-                              ^^
-  The latter is the position-1 element of the constructor "::" in the value which is the former
+Line 4, characters 34-36:
+4 |   | x :: xs as gs -> unique_ (gs, xs)
+                                      ^^
+  The latter is part of the former
```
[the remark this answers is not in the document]
riaqn: The comment is resolved because there is a newer comment on the same issue.

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -305,17 +321,18 @@ type ('a, 'b) record = { foo : 'a; bar : 'b }
...
+3 |   | (a, b) as t -> unique_ (a, t)
                                 ^
 Error: This is used uniquely here so cannot be used twice.  Another use is
-Line 2, characters 31-32:
-2 |   | (a, b) as t -> unique_ (a, t)
+Line 3, characters 31-32:
+3 |   | (a, b) as t -> unique_ (a, t)
                                    ^
-  The former is the position-0 element of the tuple which is the latter
+  The former is part of the latter
```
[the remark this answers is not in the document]
riaqn: The comment is resolved because there is a newer comment on the same issue.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1302 @@
...
+        in
+        let there_is_of_here =
+          if (Usage.extract_occurrence t').loc = here.loc then Descendant
+          else Ancestor
+        in
+        raise (MultiUse { here; there; error; there_is_of_here; path })
```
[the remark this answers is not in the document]
riaqn: The latest commit should fix this. However, I only say "Cannot use the value" instead of "Cannot use `x`", because I think the location info is already pointing at `x`.

typing/uniqueness_analysis.ml
riaqn: `t1.usage < t0.usage` is hard to define properly - I can write a conservative function but it won't detect all cases - for example, it's hard to compare two `MaybeUnique`, as they contain a list of modes.

Besides, it seems that we can factor out some common code out of `par` and `seq` and call it `Usage_tree.map2`, and `seq` doesn't enjoy the same property that you described.

typing/uniqueness_analysis.ml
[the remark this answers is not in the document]
riaqn: `map2` now redirects to `map` when one of the trees becomes empty. I think this is even better.

testsuite/tests/typing-unique/unique_analysis.ml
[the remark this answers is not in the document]
riaqn: The "part of" printing logic is (I think) quite good now. Please check.

typing/typecore.ml
```diff
@@ -4406,13 +4397,9 @@ and type_expect_
...
-      let mode = if mode_cross env ty_expected then mode_max
-      else begin
-        let mode = mode_once expected_mode in
-        submode ~loc ~env ~reason:Other mode.mode expected_mode;
-        mode
-      end
-      in
+      let expected_mode = expect_mode_cross env ty_expected expected_mode in
+      let mode = mode_once expected_mode in
+      submode ~loc ~env ~reason:Other mode.mode expected_mode;
```
lpw25, line 4402: This line is equivalent to:
```ocaml
submode ~loc ~env ~reason:Other Value.once expected_mode;
```
but much less clear. And if you reorder it with the line above then you can do:
```ocaml
let expected_mode = expect_mode_cross env ty_expected expected_mode in
submode ~loc ~env ~reason:Other Value.once expected_mode;
let expected_mode = mode_once expected_mode in
```
and avoid the need to have variables called both `mode` and `expected_mode`.

lpw25 (commented):
Reviewed all the recent changes outside of `uniqueness_analysis.ml`

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -452,13 +454,13 @@ let foo () =
...
 [%%expect{|
 type mfoo = { mutable a : string; b : string; }
-Line 11, characters 20-21:
-11 |   ignore (unique_id x);
-                         ^
-Error: This is used uniquely here so cannot be used twice.  Another use is
 Line 12, characters 2-16:
 12 |   x.a <- "olleh"
        ^^^^^^^^^^^^^^
+Error: Cannot use the value, because it is part of a value that has already been used as unique here:
```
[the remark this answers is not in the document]
riaqn: We now detect `Memory_address` and print accordingly. We can extend `Memory_address` with another field to indicate "read from" and "write to", does that sound OK?

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -420,7 +422,7 @@ let foo () =
 Line 6, characters 12-13:
 6 |   unique_id r
                 ^
-Error: This is used uniquely here so cannot be used twice.  Another use is
+Error: The value is used as unique, but part of it has already been used here:
```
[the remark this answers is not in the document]
riaqn: I guess you mean the special case where implicit borrowing in a closure is lifted to shared??

testsuite/tests/typing-unique/unique.ml
```diff
@@ -9,7 +9,7 @@ let dup x = unique_ (x, x)
 Line 1, characters 24-25:
 1 | let dup x = unique_ (x, x)
                             ^
-Error: Cannot use the value, because it has already been used as unique here:
+Error: The value is used here, but it has already been used as unique here:
```
lpw25, line 12: I think "This value" reads slightly better than "The value".

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -422,7 +422,7 @@ let foo () =
 Line 6, characters 12-13:
 6 |   unique_id r
                 ^
-Error: The value is used as unique, but part of it has already been used here:
+Error: The value is used as unique, but it has already been accessed here:
```
lpw25, line 425: This one is still confusing without the mention of a closure or function, because usually it is fine to access something before it is used. "This value is used as unique, but it is accessed within a closure here:" would be better. I think that you can probably detect this based solely on it being a shared use of the memory address, since I think that can only happen via a closure. Alternatively you could track this more directly in the analysis.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1316 @@
...
+  let set_barrier t uniq =
+    List.iter
+      (fun (barrier, _) ->
+        match !barrier with
+        | Some _ -> assert false
+        | None -> barrier := Some uniq)
+      t
+end
+
+module Usage = struct
```
lpw25, line 245: I think this module would also benefit from a signature.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1316 @@
...
+        let occ1 = Maybe_unique.extract_occurrence l1 in
+        let occ0 = Maybe_unique.extract_occurrence l0 in
+        Maybe_unique.force_shared_multiuse l0 ~there:occ1 ~first_or_second:First;
+        Maybe_unique.force_shared_multiuse l1 ~there:occ0
+          ~first_or_second:Second;
+        Shared (Maybe_unique.extract_occurrence l0)
+end
+
+(** lifting module Usage to trees *)
+module Usage_tree = struct
```
lpw25, line 421: This would benefit from a signature.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1316 @@
...
+    | [] -> { usage = leaf; children = Projection.Map.empty }
+    | proj :: path ->
+        {
+          usage = Unused;
+          children = Projection.Map.singleton proj (singleton path leaf);
+        }
+end
+
+(** Lift Usage_tree to forest *)
+module Usage_forest = struct
```
lpw25, line 524: This would benefit from a signature

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1316 @@
...
+  let singleton ((rootid, path') : Path.t) leaf =
+    Root_id.Map.singleton rootid (Usage_tree.singleton path' leaf)
+
+  (** f must be monotone *)
+  let map f = Root_id.Map.map f
+end
+
+module UF = Usage_forest
+
+module Ienv = struct
```
lpw25, line 597: This would benefit from a signature

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1316 @@
...
+      let usage = try f t.usage with exn -> on_exception exn projs in
+      let children =
+        Projection.Map.mapi (fun proj t -> loop (proj :: projs) t) t.children
+      in
+      { usage; children }
+    in
+    loop projs t
+
+  let rec map2 ?(on_exception = fun _exn _rel -> assert false) f t0 t1 =
+    {
```
lpw25, line 468: I find the style:
```ocaml
let foo = ... in
let bar = ... in
{ foo; bar }
```
to be a little easier to read than:
```ocaml
{
  foo = ...;
  bar = ... ;
}
```
when the definitions are large like this.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1316 @@
...
+    let child (p : t) (a : Projection.t) : t = p @ [ a ]
+    let root : t = []
+    let _print ppf = Format.pp_print_list Projection.print ppf
+  end
+
+  let leaf usage = { usage; children = Projection.Map.empty }
+  let _unused = leaf Usage.Unused
+
+  (** f must be monotone *)
+  let map ?(on_exception = fun _exn _projs -> assert false) ?(projs=[]) f t =
```
lpw25, line 457: The `projs` optional parameter seems to only be used in `map2`. I think it would be better to make it non-optional, call this function `map_with_projs`, define `map` in terms of it, and then only expose `map` in this module's signature.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1316 @@
...
+    let rec loop projs t =
+      let usage = try f t.usage with exn -> on_exception exn projs in
+      let children =
+        Projection.Map.mapi (fun proj t -> loop (proj :: projs) t) t.children
+      in
+      { usage; children }
+    in
+    loop projs t
+
+  let rec map2 ?(on_exception = fun _exn _rel -> assert false) f t0 t1 =
```
lpw25, line 467: As with `map`, it is easier to pass the `rel` into `f` itself rather than have a separate `on_exception` parameter.

lpw25 (commented):
Some more comments

testsuite/tests/typing-unique/unique_analysis.ml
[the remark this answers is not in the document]
lpw25: That sounds pretty easy, so I'd say go for it

testsuite/tests/typing-unique/unique_analysis.ml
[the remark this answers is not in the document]
lpw25: That's right

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1361 @@
...
+  val seq : t -> t -> t
+  (** Sequential composition lifted from [Usage.seq] *)
+
+  val par : t -> t -> t
+  (** Parallel composition lifted from [Usage.par]  *)
+
+  val singleton : Path.t -> Usage.t -> t
+  (** A singleton tree containing only one leaf *)
+
+  val map : (Path.t -> Usage.t -> Usage.t) -> t -> t
```
lpw25, line 440: I think this should have the comment about `map` only working on monotone functions.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1361 @@
...
+    | proj :: path ->
+        {
+          usage = Unused;
+          children = Projection.Map.singleton proj (singleton path leaf);
+        }
+end
+
+(** Lift Usage_tree to forest *)
+module Usage_forest : sig
+  module Path : sig
```
lpw25, line 540: I think this interface is actually easier to understand if it includes the `Root_id` (which I would just call `Root`). Then expose `Root.fresh`, `Root.fresh_of_ident` and `Path.root`, instead of `Path.fresh_root` and `Path.fresh_root_of_ident`.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1361 @@
...
+
+module UF = Usage_forest
+
+module Ienv : sig
+  type t
+
+  val par : t -> t -> t
+  val seq : t -> t -> t
+  val pars : t list -> t
+  val seqs : t list -> t
```
lpw25, line 654: I think it is a bit confusing using the `seq` and `par` names for these, since they don't quite correspond to the same thing here as they do for usages. I would suggest renaming `seq` as `concat` and renaming `par` as something like `meet`.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1361 @@
...
+    Ident.Map.merge
+      (fun _id locs0 locs1 ->
+        match (locs0, locs1) with
+        | None, None -> assert false
+        | None, Some t | Some t, None -> Some t
+        | Some locs0, Some locs1 -> Some (locs0 @ locs1))
+      ienv0 ienv1
+
+  (* ienv0 is the old env; ienv1 is probably the new bindings to be added after
+     pattern matching. ienv1 simply overwrite ienv0 *)
```
lpw25: I think this should actually fail if an ident is defined in both `ienv0` and `ienv1`, since that really shouldn't happen and we'd probably prefer to find out if it does start happening.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1361 @@
...
+  val seqs : t list -> t
+  val empty : t
+  val singleton : Ident.t -> UF.Path.t list -> t
+  val find_opt : Ident.t -> t -> UF.Path.t list option
+end = struct
+  type t = UF.Path.t list Ident.Map.t
+  (** Each identifier is mapped to a list of possible nodes, each represented by
+     a path into the forest, instead of directly ponting to the node. *)
+
+  (* used for [OR] patterns. This operation is commutative  *)
```
lpw25: Comments like this would be better placed in the signature.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1361 @@
...
+  let par ienv0 ienv1 =
+    Ident.Map.merge
+      (fun _id locs0 locs1 ->
+        match (locs0, locs1) with
+        | None, None -> assert false
+        | None, Some t | Some t, None -> Some t
```
lpw25, line 675: I think this operation should actually fail in this case. We expect every identifier bound in one branch to be bound in the other.

lpw25 (commented):
More comments

typing/uniqueness_analysis.ml
riaqn: `Texp_let` also uses `Ienv.seq` to add new bindings to the existing `ienv`, which strenghthens your point that there really should be `Ienv.Extension.t`, as they are indeed different.

typing/uniqueness_analysis.ml
riaqn: Hmm then we need to expose `UF.Path.t` as the product of `Root` and `Usage_tree.Path.t`? But I feel there is no much practical benefit in that?

typing/uniqueness_analysis.ml
riaqn: I now call them `Ienv.Extension.conjunct` and `disjunct`. Does that sound OK?

testsuite/tests/typing-unique/unique_analysis.ml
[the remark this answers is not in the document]
riaqn: This is addressed in the latest commit.

testsuite/tests/typing-unique/unique_analysis.ml
riaqn: This is addressed in the latest commit.

lpw25 (commented):
More comments

lpw25 (commented):
Reviewed updated tests

testsuite/tests/typing-unique/unique_analysis.ml
```diff
@@ -475,7 +472,8 @@ let foo () =
 Line 6, characters 12-13:
 6 |   unique_id r
                 ^
-Error: This value is used as unique, but it has already been read from here:
+Error: This value is used here as unique,
+       but it has already been captured in a closure that might be called later:
```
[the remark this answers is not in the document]
riaqn: This is now fixed.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1501 @@
...
+    loc =
+  let occ = Occurrence.mk ~access loc in
+  let mark_one path =
+    (* borrow the memory address of the parent *)
+    UF.singleton
+      (UF.Path.child path Usage_tree.Projection.Memory_address)
+      (* Currently we just generate a dummy unique_barrier ref that won't be
+         consumed. The distinction between implicit and explicit borrowing is
+         still needed because they are handled differently in closures *)
+      (Maybe_shared (Maybe_shared.singleton (ref None) occ))
```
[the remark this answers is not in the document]
riaqn: I was thinking that `mark_implicit_borrow_memory` is the current abstraction for that. I'm not sure having a function called `maybe_shared` which sounds like identical to `Maybe_shared` but secretly has `singleton`.

typing/uniqueness_analysis.ml
riaqn: I agree there is space for abstracting "list of `UF.Paths.t`" to its own type, but I'm not sure about having this inside `module UF`, as I really like the idea that any `UF.t` can be thought as a mapping from `UF.Path.t` to `Usage`.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1501 @@
...
+   - the location where it's defined
+   - the modes, if it's actually an alias
+*)
+
+type value_to_match =
+  | Match_tuple of single_value_to_match list
+      (** The value being matched is a tuple; we treat it specially so matching
+  tuples against tuples merely create alias instead of uses *)
+  | Match_single of single_value_to_match
+      (** The value being matched is not a tuple *)
```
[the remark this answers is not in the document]
riaqn: That's good insight. Fixed.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1507 @@
...
+  | Error cannot_force -> raise (Boundary { cannot_force; reason })
+
+type single_value_to_match = {
+  paths : UF.Path.t list;  (** the value's all possible paths. *)
+  loc : Location.t;  (** the location where it's defined *)
+}
+(** Represent a single value that will be pattern-matched upon *)
+
+type value_to_match =
+  | Match_tuple of (single_value_to_match * unique_use option) list
```
lpw25, line 787: I think that it would be worth lifting out `single_value_to_match * unique_use option` as a record type as well, because it should also be the return type of `check_uniqueness_exp'`. Probably we should then rename these types to be less specific to matching. How about `single_value`, `single_value_use`, and `matched_value`?

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1507 @@
...
+      (Ienv.Extension.disjunct ext0 ext1, UF.seq uf0 uf1)
+
+and pat_proj :
+      'a.
+      ?handle_tuple:_ ->
+      extract_pat:('a -> _) ->
+      mk_proj:(_ -> 'a -> _) ->
+      _ ->
+      'a list ->
+      _ =
```
lpw25, line 943: I think `pat_proj` makes things harder to understand. I think there are three changes that would simplify things and allow us to remove it:
1.  Rather than trying to get all these projection cases to behave differently on `Match_single` and `Match_tuple`, I think it would be better to have them all of them -- other than `Ppat_tuple` -- start by calling the function which turns a `value_to_match` into a `single_value_to_match`. In these cases, the type-checker should prevent us seeing a `Match_tuple` case, but if we somehow did, converting it to a single value and then projecting from that is the only thing that would make sense.
2. Lift all the projections to work on `single_value`s. That would allow lines 952-954 to get replace by a single function call in each instance.
3. Use a `fold_left`, applying `conjunct` and `seq` as we go rather than doing a `map`, a `split` and a list-based operation.

After these changes, I think you will find that the code for each case is simple enough to not need to share an auxiliary function.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1507 @@
...
+        value pats
+  | Tpat_construct (lbl, _, pats, _) ->
+      pat_proj ~extract_pat:Fun.id
+        ~mk_proj:(fun i _ ->
+          Usage_tree.Projection.Construct_field (Longident.last lbl.txt, i))
+        value pats
+  | Tpat_variant (lbl, mpat, _) ->
+      let uf = mark_implicit_borrow_memory_address Read value in
+      let t =
+        match value with
```
lpw25, line 888: This should just call the function that converts a `value_to_match` into a `single_value_to_match`.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1507 @@
...
+type value_to_match =
+  | Match_tuple of (single_value_to_match * unique_use option) list
+      (** The value being matched is a tuple; we treat it specially so matching
+  tuples against tuples merely create alias instead of uses; the [unique_use] is
+  used if the tuple is bound to a variable, in which case all values in the tuple
+    is considered used *)
+  | Match_single of single_value_to_match
+      (** The value being matched is not a tuple *)
+
+let mark_implicit_borrow_memory_address_paths access paths loc =
```
lpw25, line 795: I think you only need versions of `mark_implicit_borrow_memory_address` that work on the `single_value` and `single_value_use` types.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1507 @@
...
+  in
+  UF.pars (List.map mark_one paths)
+
+let mark_implicit_borrow_memory_address access = function
+  | Match_single { paths; loc } ->
+      mark_implicit_borrow_memory_address_paths access paths loc
+  (* it's still a tuple - we own it and nothing to do here *)
+  | Match_tuple _ -> UF.unused
+
+let mark_maybe_unique paths unique_use occ =
```
lpw25, line 815: This should also operate on `single_value_use`

lpw25 (commented):
More notes

typing/uniqueness_analysis.ml
lpw25: A single `Path.t` represents some value in our notion of "heap". It's natural to end up moving to the powerset of such a notion once there is some form of control flow involved. In our case this comes from or-patterns, but we'd get the same thing if we tried to make the analysis more precise on expressions (e.g. tracking `let x = if p then y else z in` accurately). It's no less algebraic than the rest of what is being defined here.

riaqn: We now have a `module Paths` outside of `Usage_forest` - could you take a look?

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1502 @@
...
+   usage during the process
+*)
+let pattern_match_var ~loc id value =
+  match value with
+  | Match_single (paths, _, _) ->
+      (value, Ienv.Extension.singleton id paths, UF.unused)
+  | Match_tuple values ->
+      let path = UF.Path.fresh_root_of_ident id in
+      let ext = Ienv.Extension.singleton id [ path ] in
+      (* Mark all ps as seen, as we bind the tuple to a variable. *)
```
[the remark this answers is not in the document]
riaqn: The code is refactored and comments rewritten - resolved.

typing/uniqueness_analysis.ml
riaqn: After refactoring there is 
- `Paths.t` which is a list of `Usage_forest.Path.t`
- `Value.t` which is a record containing `Paths.t`, `unique_use option` and `Occurrence.t`. This is the return type of `check_uniqueness_exp'`.
- `value_to_match` which is as usual.

Now I think that maybe the name `Paths` is too low-level compared to `single_value`, but it's also very descriptive. It's very easy to do a global search-replace.

riaqn:
After refactoring, lots of the issues pointed out by the comments are addressed. However, because of the big changes, I'm not sure if those are addressed in the ideal way. Therefore, I left those comments open. Could you take another look and let me know @lpw25

lpw25 (commented):
Some comments on recent changes. I'm also going to push some changes directly for the parts of `uniqueness_analysis.ml` I haven't yet reviewed.

typing/uniqueness_analysis.ml
```diff
@@ -540,21 +569,33 @@ module Usage_forest : sig
...
     type t
 
-    val child : t -> Projection.t -> t
-    val child_of_many : t list -> Projection.t -> t list
-    val fresh_root_of_ident : Ident.t -> t
+    val child : Usage_tree.Projection.t -> t -> t
+    (** Construct a child path from a parent  *)
+
     val fresh_root : string -> t
+    (** Create a fresh tree in the forest, using the string as a hint name *)
```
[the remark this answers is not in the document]
riaqn: I think it's only used for debugging. Should I remove it?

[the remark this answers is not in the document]
lpw25: Yeah

typing/uniqueness_analysis.ml
```diff
@@ -619,258 +654,309 @@ end = struct
...
+  let ufs = match prologue with None -> ufs | Some uf -> uf :: ufs in
+  (Ienv.Extension.conjuncts exts, UF.seqs ufs)
+
+let rec pattern_match pat vtm =
+  match (vtm, pat.pat_desc) with
+  | _, Tpat_or (a, b, _) ->
+      let ext0, uf0 = pattern_match a vtm in
+      let ext1, uf1 = pattern_match b vtm in
+      (Ienv.Extension.disjunct ext0 ext1, UF.seq uf0 uf1)
+  | _, Tpat_lazy pat -> pattern_match pat vtm
```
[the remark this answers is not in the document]
riaqn: Yeah that sounds reasonable. For this reason I think `Match_single` also needs to contain a `Value.t` instead of just `Paths.t`; shouldn't be hard to change.

typing/typecore.ml
```diff
@@ -4198,42 +4372,64 @@ and type_expect_
...
-        (* moreover, because mode crosses, expected_mode is completely useless *)
-      else begin
-        (* if mode does not cross, expected.mode must be local *)
-        submode ~loc ~env ~reason:Other Value_mode.local expected_mode;
-        (* and we require the inner expr to be exact local *)
-        {expected_mode with mode = Value_mode.local; exact = true}
-      end
+      if txt = "extension.unique" && not (Language_extension.is_enabled Unique) then
+        raise (Typetexp.Error (loc, Env.empty, Unsupported_extension Unique));
+      let expected_mode = mode_unique expected_mode in
```
riaqn, line 4388: I now want to add back the line `submode unique expected_mode` before this line. It is true that this would be a no-op, but `typecore.ml` doesn't need to know that `unique` the least mode in the axis? I think currently `typecore.ml` doesn't assume `global` is min and `local` is max.

lpw25: It's not about it being the least mode it's about the meaning of `unique_ e`. It is a request to ensure that the `e` is at least `unique_`. Consider what `exclusive_ e` should mean when the result of `e` is `unique` and is then used uniquely. I would say that should not be an error: `exclusive_ e` is a request to ensure that the result of `e` is at least `exclusive` not a request that the result of the expression be at most `exclusive.

riaqn: > It is a request to ensure that the e is at least unique_

I guess you mean "e is stricter than `unique`"? But in any case, the original `expected_mode` shouldn't just be ignored (for soundness), except here because `unique` is the lowest mode?

lpw25: Right, `mode_unique expected_mode` should return the meet of `expected_mode` and `unique`, which is always `unique` because `unique` is the lowest mode.

riaqn: But with the same logic, `once_ e` would be useless? Because `once` is already the most relaxed mode in that axis. Or does it mean "relax `e` to `once` and error if later used as `many`"?

lpw25: `once_` is interpreted in the other direction, as a requirement on the `expected_mode`. My point is that I think we should choose to interpret different modes differently when used as expressions. I think the distinction is where the mode sits in its axis relative to the legacy mode. If a mode is stricter that legacy then it should be interpreted as a requirement on the expression, if a mode is less strict than legacy then it should be interpreted as a requirement on the context.

testsuite/tests/typing-unique/unique.ml
```diff
@@ -544,7 +541,7 @@ let curry : unique_ box -> (unique_ box -> unique_ box) = fun b1 b2 -> b1
 Line 1, characters 58-73:
 1 | let curry : unique_ box -> (unique_ box -> unique_ box) = fun b1 b2 -> b1
                                                               ^^^^^^^^^^^^^^^
-Error: This function or one of its parameters will be at once but expected to be many
+Error: This function or one of its parameters will be once but will be used many
```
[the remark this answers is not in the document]
riaqn: I have incorporated the first option because it's probably better to treat the `linearity` case of `Uncurried_function_escape` in the same way as the other axes, and for those axes we don't print the whole type signature.

typing/uniqueness_analysis.ml
```diff
@@ -0,0 +1,1502 @@
...
+    UF.singleton path (Maybe_unique (Maybe_unique.singleton unique_use occ))
+  in
+  UF.pars (List.map (fun path -> mark_one path) paths)
+
+(* returns:
+   the updated value.
+   the new introduced bindings.
+   usage during the process
+*)
+let pattern_match_var ~loc id value =
```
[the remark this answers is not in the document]
riaqn: Resolved because the code no longer exists.

typing/uniqueness_analysis.ml
riaqn: I believe this is considered resolved.

typing/uniqueness_analysis.ml
riaqn: I believe this is considered resolved.

typing/uniqueness_analysis.ml
riaqn: Resolved because we have taken a slight different approach.

typing/uniqueness_analysis.ml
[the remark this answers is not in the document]
riaqn: In the end I only added the one for `Usage.shared`, because this is the only one that's would be used more than once and thus worth adding a shorthand for.

typing/uniqueness_analysis.ml
```diff
@@ -102,13 +105,13 @@ module Maybe_shared : sig
...
+  (** set a barrier. The uniqueness mode represents the usage immediately
       following the current usage. If that mode is Unique, the current usage
        must be Borrowed (hence no code motion); if that mode is not restricted
-       to Unique, this usage can be Borrowed or Shared (prefered). Can be called
-       only once; raises if set multiple times *)
+       to Unique, this usage can be Borrowed or Shared (prefered). Raise if
+        called more than once. *)
 
-  val par : t -> t -> t
+  val choose : t -> t -> t
```
lpw25, line 114: As discussed offline, this should be called `meet`.

typing/uniqueness_analysis.ml
```diff
@@ -43,16 +43,16 @@ module Maybe_unique : sig
   val singleton : unique_use -> Occurrence.t -> t
   (** construct a single usage *)
 
-  val par : t -> t -> t
+  val choose : t -> t -> t
```
lpw25, line 46: This should be called `meet`.

typing/uniqueness_analysis.ml
```diff
@@ -381,94 +496,32 @@ module Usage_tree : sig
...
-
   val seq : t -> t -> t
   (** Sequential composition lifted from [Usage.seq] *)
 
+  val choose : t -> t -> t
+  (** non-det choice lifted from [Usage.choose] *)
+
   val par : t -> t -> t
-  (** Parallel composition lifted from [Usage.par]  *)
+  (** parallel composition lifted from [Usage.par]  *)
```
lpw25, line 506:
```suggestion
  (** Parallel composition lifted from [Usage.par]  *)
```

typing/uniqueness_analysis.ml
```diff
@@ -381,94 +496,32 @@ module Usage_tree : sig
...
-        (** The relation between the two usages in the tree  *)
-  }
-
-  exception Error of error
-
   val seq : t -> t -> t
   (** Sequential composition lifted from [Usage.seq] *)
 
+  val choose : t -> t -> t
+  (** non-det choice lifted from [Usage.choose] *)
```
lpw25, line 503:
```suggestion
  (** Non-deterministic choice lifted from [Usage.choose] *)
```

typing/uniqueness_analysis.ml
```diff
@@ -284,17 +300,51 @@ end = struct
...
+    | Borrowed occ, Borrowed _ -> Borrowed occ
+    | Borrowed _, Maybe_shared t | Maybe_shared t, Borrowed _ -> Maybe_shared t
+    | Borrowed _, Shared t | Shared t, Borrowed _ -> Shared t
+    | Borrowed occ, Maybe_unique t | Maybe_unique t, Borrowed occ ->
+        force_shared_multiuse t (Borrowed occ) First;
+        shared (Maybe_unique.extract_occurrence t) Shared.Forced
+    | Maybe_shared t0, Maybe_shared t1 ->
+        Maybe_shared (Maybe_shared.choose t0 t1)
+    | Maybe_shared _, Shared occ | Shared occ, Maybe_shared _ ->
+        (* the barrier stays empty; if there is any unique after this, it
```
lpw25, line 320:
```suggestion
        (* The barrier stays empty; if there is any unique after this, it
```

typing/uniqueness_analysis.ml
```diff
@@ -195,8 +201,11 @@ module Usage : sig
   val seq : t -> t -> t
   (** Sequential composition *)
 
+  val choose : t -> t -> t
+  (** non-det choice *)
+
   val par : t -> t -> t
-  (** Parallel composition  *)
+  (** par composition *)
```
lpw25, line 208:
```suggestion
  (** Parallel composition *)
```

typing/uniqueness_analysis.ml
```diff
@@ -381,94 +496,32 @@ module Usage_tree : sig
...
      INVARIANT: children >= parent. For example, having a shared child under a
-     unique parent is nonsense. The invariant is preserved because Usage.par and
-     Usage.seq above is monotone, and Usage_tree.par and Usage_tree.seq here is
-     node-wise. *)
-
-  type error = { inner : Usage.error; first_is_of_second : relation }
-
-  exception Error of error
+     unique parent is nonsense. The invariant is preserved because Usage.choose,
+     Usage.par, and Usage.seq above is monotone, and Usage_tree.par and
```
lpw25, line 523:
```suggestion
     Usage.par, and Usage.seq above are monotone, and Usage_tree.par and
```

typing/uniqueness_analysis.ml
```diff
@@ -381,94 +496,32 @@ module Usage_tree : sig
...
+     Usage_tree.seq, Usage_tree.choose here is node-wise. *)
```
lpw25, line 524:
```suggestion
     Usage_tree.seq, Usage_tree.choose here are node-wise. *)
```

typing/uniqueness_analysis.ml
```diff
@@ -514,16 +564,18 @@ end = struct
...
     mapi2
       (fun first_is_of_second t0 t1 ->
-        try Usage.seq t0 t1
+        try f t0 t1
         with Usage.Error error ->
-          raise (Error { inner = error; first_is_of_second }))
+          raise (Error (Usage { inner = error; first_is_of_second })))
       t0 t1
 
+  let choose = lift Usage.choose
```
lpw25, line 575:
```suggestion
  let choose t1 t2 = lift Usage.choose t1 t2
```

typing/uniqueness_analysis.ml
```diff
@@ -514,16 +564,18 @@ end = struct
...
+  let seq = lift Usage.seq
+  let par = lift Usage.par
```
lpw25, line 577:
```suggestion
  let par t1 t2 = lift Usage.par t1 t2
```

lpw25 (approved):
Some final suggestions, but I think this is probably good to go.

PR #139, opened 2023-03-08
Unboxed types v1

ccasin:
This is the first unboxed types release.  It adds layout inference and a very limited collection of layouts.  See the unboxed types proposal (for example, [the rfc here](https://github.com/ocaml/RFCs)) for details on the long-term plan.


goldfirere:
CI is now green here. If you submit a patch that doesn't pass CI, that means you've introduced a problem that's worth fixing.

lambda/matching.ml
```diff
@@ -2790,41 +2828,58 @@ let combine_constant value_kind loc arg cst partial ctx def
...
+        | Extension _, _ -> assert false
       )
   in
   let const, nonconst = split_rec tag_lambda_list in
   (sort_int_lambda_list const, sort_int_lambda_list nonconst)
 
+(* The bool tracks whether the constructor is constant, because we don't have a
+   constructor_description available for polymorphic variants *)
+let split_variant_cases (tag_lambda_list : ((int * bool) * lambda) list) =
+  let rec split_rec = function
```
goldfirere, line 2849: I think this is just `List.partition_map ( fun ((n, is_const), act) -> if is_const then Left (n, act) else Right (n, act) )`.

lambda/matching.ml
```diff
@@ -2790,41 +2828,58 @@ let combine_constant value_kind loc arg cst partial ctx def
...
+  in
+  let const, nonconst = split_rec tag_lambda_list in
+  (sort_int_lambda_list const, sort_int_lambda_list nonconst)
+
+
 let split_extension_cases tag_lambda_list =
   let rec split_rec = function
     | [] -> ([], [])
-    | (cstr_tag, act) :: rem -> (
+    | ({cstr_constant; cstr_tag}, act) :: rem -> (
```
goldfirere, line 2865: This also looks like a `partition_map`.

lambda/matching.mli
```diff
@@ -19,6 +19,11 @@ open Typedtree
 open Lambda
 open Debuginfo.Scoped_location
 
+type error =
```
goldfirere, line 22: Do these declarations need to be here? I think not.

lambda/translmod.ml
```diff
@@ -1582,13 +1596,15 @@ let close_toplevel_term (lam, ()) =
 
 let transl_toplevel_item ~scopes item =
   match item.str_desc with
-    Tstr_eval (expr, _)
+    (* These first two cases are special compilation for toplevel "let _ =
+       expr", so that Toploop can display the result of the expression.
+       Otherwise, the normal compilation would result in a Lsequence returning
+       unit. *)
+    Tstr_eval (expr, layout, _) ->
+      layout_must_be_value expr.exp_loc expr.exp_type layout;
```
goldfirere, line 1604: Does this really need to be value? I would expect a "make sure it's not `void`" check here, in line with the check for `;` expressions

lambda/translmod.ml
```diff
@@ -1582,13 +1596,15 @@ let close_toplevel_term (lam, ()) =
...
+      transl_exp ~scopes expr
   | Tstr_value(Nonrecursive,
                [{vb_pat = {pat_desc=Tpat_any};vb_expr = expr}]) ->
```
goldfirere, line 1607: Should there be a sort check here?

goldfirere (commented):
Reviewed the transl files. Will review Typeopt later.

lambda/translcore.ml
```diff
@@ -1609,6 +1735,14 @@ let report_error ppf = function
         "Ancestor names can only be used to select inherited methods"
   | Unreachable_reached ->
       fprintf ppf "Unreachable expression was reached"
+  | Bad_probe_layout id ->
+      fprintf ppf "Variables in probe handlers must have layout value, \
+                   but %s in this handler does not." (Ident.name id)
```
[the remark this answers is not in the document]
ccasin: At the moment you'd have to be doing something very experimental to hit this (you need a non-value), so I think it's fine?  I'll leave a v2 CR on the probe hack to re-evaluate as we get closer to letting somebody actually reach this error.

lambda/matching.ml
ccasin: I've changed these.  I hesitated to do so for two reasons:
1) Minimizing the diff in this patch to the relevant bits (here, `split_extension_cases` is a pre-existing function with this logic, and the other is meant to mirror it.  But clearly that ship has sailed in this PR.
2) Now we're allocating extra "Either"s.

But in the end I agree it's an improvement.

lambda/translmod.ml
ccasin: Good point!  As a result, transl was incorrectly rejecting:

```
module M = struct
  assert false
end
```

I've fixed this (and the two other places where `Tstr_eval` is handled)

lambda/matching.mli
goldfirere: In every case I've looked at, the types are in the mli file because some other module needs to get at them. That's not the case here.

ccasin: OK, removed.

typing/typeopt.ml
```diff
@@ -179,194 +189,330 @@ let bigarray_type_kind_and_layout env typ =
...
+              | Mutable -> true
+              | Immutable -> is_mutable
+            in
+            is_mutable, num_nodes_visited, idx+1, kinds)
+          (false, num_nodes_visited, 0, []) labels
+      in
+      let kinds = List.rev kinds in
+      match is_mutable, kinds with
+      | true, _ -> (num_nodes_visited, Pgenval)
+      | false, [] -> (num_nodes_visited, Pintval)
```
goldfirere, line 478: I don't see this case in the old code... unless it's the `gc_always_ignorable` check (which it probably is).

typing/typeopt.ml
```diff
@@ -179,194 +189,330 @@ let bigarray_type_kind_and_layout env typ =
...
+  | Variant_boxed _layouts ->
+    let depth = depth + 1 in
+    let for_one_constructor (constructor : Types.constructor_declaration)
+          ~depth ~num_nodes_visited =
+      let num_nodes_visited = num_nodes_visited + 1 in
+      match constructor.cd_args with
+      | Cstr_tuple fields ->
+        let num_nodes_visited, _, kinds =
+          List.fold_left
+            (fun (num_nodes_visited, idx, kinds) (ty,_) ->
```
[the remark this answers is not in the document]
ccasin: We spoke in person so I'll just comment on one of these for future reference:  All the style nits / unnecessary changes in this file were part of the previous mechanism for handling void.  I've removed them.

goldfirere (approved):

goldfirere:
:tada: :raised_hands: :shipit:

PR #140, opened 2023-03-08
Some small patch-ups around matching on extensions

goldfirere:
EDIT: This was originally a much more ambitious patch, but has been scaled down because the original design would make upstreaming harder.

------------------------------------

The goal here is to more structurally avoid the possibility of
forgetting about extensions. This doesn't really actually get us
there, though, because we can't change Parsetree to avoid using
e.g. pexp_desc. And even if we could make that change, we might
not want to, because it's just too convenient to just use pexp_desc.
Until we have active patterns of some sort, we may be unable to
improve upon the status quo.

However, as I was working this, I identified a few spots where I
think there are live bugs for lack of looking for extensions. And
I think having a destructor, as this commit adds, is a positive little
improvement.

This PR is against the `rae/no-raw-type` branch, because I'm stacking PRs.

### DIFF
--- a/parsing/extensions.mli
+++ b/parsing/extensions.mli
@@ -97,7 +97,12 @@ module type AST = sig
       if it's not a language extension term, return [None]; if it's a disabled
       language extension term, raise an error.
 
-      AN IMPORTANT NOTE: We indent calls to this function *very* strangely: we
+      AN IMPORTANT NOTE: The design of this function is careful to make
+      merge conflicts with upstream less likely: we want no edits at all
+      to surrounding code. This is why we return a [t option], not some
+      structure that could include the [ast_desc] if there is no extension.
+
+      In addition, we indent calls to this function *very* strangely: we
       *do not change the indentation level* when we match on its result!
       E.g. from [type_expect_] in [typecore.ml]:
 
@@ -126,9 +131,7 @@ module type AST = sig
       Note that we match on the result of this function, forward to
       [type_expect_extension] if we get something, and otherwise do the real
       match on [sexp.pexp_desc] *without going up an indentation level*.  This
-      is important to reduce the number of merge conflicts with upstream by
-      avoiding changing the body of every single important function in the type
-      checker to add pointless indentation. *)
+      is important to reduce the number of merge conflicts. *)
   val of_ast : ast -> t option
 end
 
--- a/parsing/extensions.mli
+++ b/parsing/extensions.mli
@@ -17,6 +17,9 @@
     For details on the rationale behind this approach (and for some of the gory
     details), see [Extensions_parsing]. *)
 
+(*********************************************)
+(* Individual extensions *)
+
 (** The ASTs for list and array comprehensions *)
 module Comprehensions : sig
   type iterator =
@@ -76,58 +79,8 @@ module Immutable_arrays : sig
   val pat_of : loc:Location.t -> pattern -> Parsetree.pattern_desc
 end
 
-(** The module type of language extension ASTs, instantiated once for each
-    syntactic category.  We tend to call the pattern-matching functions here
-    with unusual indentation, not indenting the [None] branch further so as to
-    avoid merge conflicts with upstream. *)
-module type AST = sig
-  (** The AST for all our ocaml-jst language extensions; one constructor per
-      language extension that extends the expression language.  Some extensions
-      are handled separately and thus are not listed here. *)
-  type t
-
-  (** The corresponding OCaml AST *)
-  type ast
-
-  (** Given an OCaml AST node, check to see if it corresponds to a language
-      extension term.  If it is, and the extension is enabled, then return it;
-      if it's not a language extension term, return [None]; if it's a disabled
-      language extension term, raise an error.
-
-      AN IMPORTANT NOTE: We indent calls to this function *very* strangely: we
-      *do not change the indentation level* when we match on its result!
-      E.g. from [type_expect_] in [typecore.ml]:
-
-      {[
-        match Extensions.Expression.of_ast sexp with
-        | Some eexp ->
-            type_expect_extension
-              ~loc ~env ~expected_mode ~ty_expected ~explanation eexp
-        | None      -> match sexp.pexp_desc with
-        | Pexp_ident lid ->
-            let path, mode, desc, kind = type_ident env ~recarg lid in
-            (* ... *)
-        | Pexp_constant(Pconst_string (str, _, _) as cst) ->
-            register_allocation expected_mode;
-            (* ... *)
-        | (* ... *)
-        | Pexp_unreachable ->
-            re { exp_desc = Texp_unreachable;
-                 exp_loc = loc; exp_extra = [];
-                 exp_type = instance ty_expected;
-                 exp_mode = expected_mode.mode;
-                 exp_attributes = sexp.pexp_attributes;
-                 exp_env = env }
-      ]}
-
-      Note that we match on the result of this function, forward to
-      [type_expect_extension] if we get something, and otherwise do the real
-      match on [sexp.pexp_desc] *without going up an indentation level*.  This
-      is important to reduce the number of merge conflicts with upstream by
-      avoiding changing the body of every single important function in the type
-      checker to add pointless indentation. *)
-  val of_ast : ast -> t option
-end
+(******************************************)
+(* Individual syntactic categories *)
 
 (** Language extensions in expressions *)
 module Expression : sig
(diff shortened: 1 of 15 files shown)


parsing/extensions.mli
```diff
@@ -76,72 +79,28 @@ module Immutable_arrays : sig
   val pat_of : loc:Location.t -> pattern -> Parsetree.pattern_desc
 end
 
-(** The module type of language extension ASTs, instantiated once for each
-    syntactic category.  We tend to call the pattern-matching functions here
-    with unusual indentation, not indenting the [None] branch further so as to
-    avoid merge conflicts with upstream. *)
```
antalsz, line 82: I think we lost the comments about indentation, and they're very important documentation

parsing/extensions.mli
```diff
@@ -76,72 +79,28 @@ module Immutable_arrays : sig
...
-                 exp_attributes = sexp.pexp_attributes;
-                 exp_env = env }
-      ]}
-
-      Note that we match on the result of this function, forward to
-      [type_expect_extension] if we get something, and otherwise do the real
-      match on [sexp.pexp_desc] *without going up an indentation level*.  This
-      is important to reduce the number of merge conflicts with upstream by
-      avoiding changing the body of every single important function in the type
-      checker to add pointless indentation. *)
```
antalsz, line 128: I think we lost the comments about indentation, and they're very important documentation.  Could this comment or something like it move on to `get_desc`?

parsing/parsetree.mli
```diff
@@ -211,6 +211,12 @@ and object_field_desc =
 and pattern =
     {
      ppat_desc: pattern_desc;
+     (** Prefer [Extensions.Pattern.get_desc] instead of using
+         this field directly, as the former will detect extension nodes
+         correctly. Extensions are encoded as
+         [Ppat_tuple [Ppat_extension _; _]]; if your pattern match avoids
+         matching that pattern, it is OK to skip [get_desc]. *)
+
```
antalsz, line 219: Do we want to mark this comment (and the analogous later one) as being `ocaml-jst` specific?

parsing/pprintast.ml
```diff
@@ -66,6 +66,8 @@ let fixity_of_string  = function
   | _ -> `Normal
 
 let view_fixity_of_exp = function
+  (* matching on pexp_desc is OK because we're looking for Pexp_ident,
+     which is different from the Pexp_apply node of an extension *)
```
antalsz, line 70: I don't love how fiddly these comments are, and how they break the abstraction barrier.  They may be worth it, but I wanted to register my concern.

parsing/pprintast.ml
```diff
@@ -461,16 +465,18 @@ and pattern_or ctxt f x =
...
-      ->
+            Some ([], inner_pat));
+       ppat_attributes = []} ->
+      begin match Extensions.Pattern.get_desc inner_pat with
+      | Regular(Ppat_tuple([pat1; pat2])) ->
         pp f "%a::%a" (simple_pattern ctxt) pat1 pattern_list_helper pat2 (*RA*)
-    | p -> pattern1 ctxt f p
+      | _ -> pattern1 ctxt f p
+      end
+    | _ -> pattern1 ctxt f p
```
antalsz, line 479: This is a good relatively extremal example of the sort of diff noise I was worried about

antalsz (commented):
I have mixed feelings about this.  On the one hand, I like it a lot -- saying "use `get_desc` instead of `.foo_desc`" is genuinely great.  On the other hand, I worry about diffs with upstream.  One careful design decision around `of_ast` was that we didn't need to edit the existing matches _at all_.  We now have to change function preambles in terms of what they bring in to scope, edit what matches match on, and so forth.  These may mostly be resolvable automatically, but I have some concerns about that.

parsing/pprintast.ml
```diff
@@ -846,7 +855,10 @@ and expression2 ctxt f x =
 
 and simple_expr ctxt f x =
   if x.pexp_attributes <> [] then expression ctxt f x
-  else match x.pexp_desc with
+  else match Extensions.Expression.get_desc x with
+    | Extension _eext -> paren true (expression ctxt) f x
```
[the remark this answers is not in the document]
goldfirere: I've reverted this, but I think there is a bug. I added a ticket to the Epic for the next version of the comprehensions patch.

goldfirere:
You're right. I've mostly reverted. I didn't quite catch the main payload of the big comments as "this avoids merge conflicts". I've revised to make this clearer. Please re-review. Thanks!

goldfirere:
(After this is reviewed, I'll rebase the rest of the chain.)

parsing/extensions.mli
```diff
@@ -97,7 +97,12 @@ module type AST = sig
       if it's not a language extension term, return [None]; if it's a disabled
       language extension term, raise an error.
 
-      AN IMPORTANT NOTE: We indent calls to this function *very* strangely: we
+      AN IMPORTANT NOTE: The design of this function is careful to make
+      merge conflicts with upstream less likely: we want no edits at all
+      to surrounding code. This is why we return a [t option], not some
+      structure that could include the [ast_desc] if there is no extension.
+
+      In addition, we indent calls to this function *very* strangely: we
```
antalsz, line 105: I'd like to prioritize the note about indentation, since that's the first thing *users* of the function will note.  I tried moving your documentation to the bottom and putting it in its own paragraph; see the other suggestion.

```suggestion
      AN IMPORTANT NOTE: We indent calls to this function *very* strangely: we
```

parsing/extensions.mli
```diff
@@ -126,9 +131,7 @@ module type AST = sig
       Note that we match on the result of this function, forward to
       [type_expect_extension] if we get something, and otherwise do the real
       match on [sexp.pexp_desc] *without going up an indentation level*.  This
-      is important to reduce the number of merge conflicts with upstream by
-      avoiding changing the body of every single important function in the type
-      checker to add pointless indentation. *)
+      is important to reduce the number of merge conflicts. *)
```
antalsz, line 134:
```
      match on [sexp.pexp_desc] *without going up an indentation level*.
      
      This design, and the design of the function in general, is done very
      carefully in order to make merge conflicts with upstream less likely: we
      want its uses to require no edits at all to surrounding code.  This
      explains the indentation; it's also why, for instance, it returns a [t
      option], not some structure that could include the [ast_desc] if there
      weren't an extension. *)
```

(GitHub won't let me make this an official suggestion.)

parsing/pprintast.ml
```diff
@@ -461,16 +461,18 @@ and pattern_or ctxt f x =
...
-      ->
+            Some ([], inner_pat));
+       ppat_attributes = []} ->
+      begin match Extensions.Pattern.of_ast inner_pat, inner_pat.ppat_desc with
+      | None, Ppat_tuple([pat1; pat2]) ->
         pp f "%a::%a" (simple_pattern ctxt) pat1 pattern_list_helper pat2 (*RA*)
-    | p -> pattern1 ctxt f p
+      | _ -> pattern1 ctxt f p
+      end
+    | _ -> pattern1 ctxt f p
```
antalsz, line 475: Is there really no way to shrink this diff?  (If not, feel free to just mark this as resolved; that happens.)

typing/typecore.ml
```diff
@@ -4068,8 +4068,8 @@ and type_expect_
         rt, funct
       in
       let type_sfunct_args sfunct extra_args =
-        match sfunct.pexp_desc with
-        | Pexp_apply (sfunct, args) ->
+        match Extensions.Expression.of_ast sfunct, sfunct.pexp_desc with
+        | None, Pexp_apply (sfunct, args) ->
```
antalsz, line 4072: This is a good idea for minimal diff noise when you want to share a default case!  I like it.  (Mark as resolved when read.)

antalsz (changes requested):
Thanks, Richard, trying to make the comment highlight that more is a good plan.  I think this is mostly good now, I just had a couple minor notes around comment wording.  We should also rename this PR now that what it does is different.

I think you made the right call by reverting, but it's not obvious; there were wins in the `get_desc` design!

parsing/pprintast.ml
goldfirere: I mean, I could make a tiny little syntactic refactoring patch to push upstream and then have the diff here be smaller, but I don't think it's worth it. For what it's worth, I think this change is likely fixing a proper live bug. I will mark as resolved, but I acknowledge that this work will make the next rebase harder.

parsing/extensions.mli
goldfirere: I'm going to mark this as resolved, but we can keep iterating in another PR if there are further improvements.

PR #141, opened 2023-03-08
Add support for extensions in module types

goldfirere:
This doesn't actually support strengthening, but just lays the groundwork for doing so. For strengthening support, wait for the next PR, coming up soon.


antalsz (approved):
LGTM

PR #142, opened 2023-03-09
Add a Strengthen extension

goldfirere:
This should set the stage for a rebase of #119. Note that there are 4 new `failwith`s I added for @rleshchinskiy to fill in.

(This has not been adequately tested, as I'm on my personal computer on a train. But CI should fill in that gap. I know the code compiles and passes most tests.)

### DIFF
--- a/parsing/ast_iterator.ml
+++ b/parsing/ast_iterator.ml
@@ -304,6 +304,12 @@ module MT = struct
         sub.attributes sub attrs;
         sub.extension sub x
     | Psig_attribute x -> sub.attribute sub x
+
+  let iter_extension sub : Extensions.Module_type.t -> _ = function
+    | Emty_strengthen { mty; mod_id } ->
+       iter sub mty;
+       iter_loc sub mod_id
+
 end
 
 
@@ -597,6 +603,7 @@ let default_iterator =
     signature = (fun this l -> List.iter (this.signature_item this) l);
     signature_item = MT.iter_signature_item;
     module_type = MT.iter;
+    module_type_extension = MT.iter_extension;
     with_constraint = MT.iter_with_constraint;
     class_declaration =
       (fun this -> CE.class_infos this (this.class_expr this));
@@ -656,9 +663,6 @@ let default_iterator =
          this.location this pmtd_loc;
          this.attributes this pmtd_attributes;
       );
-
-    module_type_extension = (fun _this emty -> match emty with
-      | _ -> .);
 
     module_binding =
       (fun this {pmb_name; pmb_expr; pmb_attributes; pmb_loc} ->
--- a/parsing/ast_iterator.ml
+++ b/parsing/ast_iterator.ml
@@ -309,7 +309,6 @@ module MT = struct
     | Emty_strengthen { mty; mod_id } ->
        iter sub mty;
        iter_loc sub mod_id
-
 end
 
 
--- a/parsing/depend.ml
+++ b/parsing/depend.ml
@@ -341,7 +341,7 @@ and add_binding_op bv bv' pbop =
 
 and add_modtype bv mty =
   match Extensions.Module_type.of_ast mty with
-  | Some _ -> .
+  | Some emty -> add_modtype_extension bv emty
   | None ->
   match mty.pmty_desc with
     Pmty_ident l -> add bv l
@@ -373,6 +373,11 @@ and add_modtype bv mty =
   | Pmty_typeof m -> add_module_expr bv m
   | Pmty_extension e -> handle_extension e
 
+and add_modtype_extension bv : Extensions.Module_type.t -> _ = function
+  | Emty_strengthen { mty; mod_id } ->
+     add_modtype bv mty;
+     add_module_path bv mod_id
+
 and add_module_alias bv l =
   (* If we are in delayed dependencies mode, we delay the dependencies
        induced by "Lident s" *)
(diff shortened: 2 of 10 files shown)


parsing/extensions.ml
```diff
@@ -297,7 +297,36 @@ module Immutable_arrays = struct
...
+         functor (_ : [%extension.strengthen]) ->
+           functor (_ : S) -> M
+       ]}
+  *)
+
+  let mty_of ~loc { mty; mod_id } =
+    (* See Note [Wrapping with make_extension] *)
+    Module_type.make_extension ~loc [extension_string] @@
+      Ast_helper.Mty.functor_ (Named (Location.mknoloc None, mty))
+        (Ast_helper.Mty.ident mod_id)
```
goldfirere, line 324: This is a bit suspect, because it's stashing a *module* identifier where parsetree expects a *module type* identifier. All of the extensions packing and unpacking code is fine with this. But could a ppx take a pass looking for all module identifiers and then miss this one? I'm starting to doubt my choice of encoding around this point.

parsing/ast_iterator.ml
```diff
@@ -303,6 +303,12 @@ module MT = struct
         sub.attributes sub attrs;
         sub.extension sub x
     | Psig_attribute x -> sub.attribute sub x
+
+  let iter_extension sub : Extensions.Module_type.t -> _ = function
+    | Emty_strengthen { mty; mod_id } ->
+       iter sub mty;
+       iter_loc sub mod_id
+
```
antalsz, line 311:
```suggestion
```

(stray newline)

parsing/depend.ml
```diff
@@ -398,6 +403,13 @@ and add_modtype_binding bv mty =
   | _ ->
       add_modtype bv mty; bound
 
+and add_modtype_extension_binding bv : Extensions.Module_type.t -> _ = function
+  | Emty_strengthen { mty; mod_id } ->
+     (* treat like a [with] constraint *)
+     add_modtype bv mty;
+     add_module_path bv mod_id;
+     bound
```
antalsz, line 411: Are you sure Roman doesn't need to implement this?

parsing/extensions.ml
```diff
@@ -297,7 +297,36 @@ module Immutable_arrays = struct
 
   let of_pat expr = match expr.ppat_desc with
     | Ppat_array elts -> Iapat_immutable_array elts
-    | _ -> failwith "Malformed immutable array expression"
+    | _ -> failwith "Malformed immutable array pattern"
```
antalsz, line 300: Thanks for catching that :-)

parsing/extensions.ml
```diff
@@ -297,7 +297,36 @@ module Immutable_arrays = struct
...
+  let extension_string = Language_extension.to_string Strengthen
+
+  (* Encoding: [S with M] becomes [functor (_ : S) -> M]. But remember
+     that this gets wrapped in an [[%extension.strengthen]] node, so
+     the what starts as [S with M] really becomes
+
+       {[
+         functor (_ : [%extension.strengthen]) ->
+           functor (_ : S) -> M
+       ]}
```
antalsz, line 317: I don't think this part of the comment belongs here; a user not reading `extensions_parsing.ml` should never have to think about the encoding we choose.  See the pseudo-BNF I have for comprehensions.

parsing/extensions.ml
antalsz: What about `S with M` becoming

```
sig
  include S
  include module type of M
end
```

?  That seems like the closest analog to a pair we have, and at a quick glance `module type of` is the only way we have to reference the module language from the module type language without going via first-class modules.

utils/language_extension.ml
```diff
@@ -27,13 +29,15 @@ let to_string = function
   | Include_functor -> "include_functor"
   | Polymorphic_parameters -> "polymorphic_parameters"
   | Immutable_arrays -> "immutable_arrays_experimental"
+  | Strengthen -> "strengthen"
```
antalsz, line 32: :bike::hut:: What about `Module_strengthening`?  I find "strengthen" lacking too much context.

antalsz (changes requested):
Mostly looks good, some small notes.

parsing/extensions.ml
rleshchinskiy: I'd replace `Pmty_ident` with `Pmty_alias` in the original encoding or go with @antalsz's suggestion (perhaps again, using `Pmty_alias` instead of `Pmty_typeof`).

utils/language_extension.ml
goldfirere: (I spent several seconds wondering what "motorbike-hut" meant, as that's what originally parsed.)

Yes, you're right.

goldfirere: Actually, I think @rleshchinskiy gets the final call here. (The pedant in me prefers `module_type_strengthening`, though that's really too long.) But I've made this change in this patch.

goldfirere:
OK. Does anyone want to click Approve?

parsing/extensions.ml
```diff
@@ -296,7 +296,30 @@ module Immutable_arrays = struct
...
+(** Module strengthening *)
+module Strengthen = struct
+  type nonrec module_type =
+    { mty : Parsetree.module_type; mod_id : Longident.t Location.loc }
+
+  let extension_string = Language_extension.to_string Module_strengthening
+
+  (* Encoding: [S with M] becomes [functor (_ : S) -> (module M)], where
+     the [(module M)] is a [Pmty_alias].
+  *)
```
antalsz, line 311:
```suggestion
  (* Encoding: [S with M] becomes [functor (_ : S) -> (module M)], where
     the [(module M)] is a [Pmty_alias].  This isn't syntax we can write, but
     [(module M)] can be the inferred type for [M], so this should be fine. *)
```

antalsz (approved):
After talking to @goldfirere I now understand the new encoding of `S with M`; I want to update the comment about it but we can merge besides that!

PR #144, opened 2023-03-09
Add `promote-failed` targets to the Makefiles

antalsz:
This target promotes all tests that failed on the previous run; it's good for after running `make -C testsuite parallel`, for example.

If this looks good here I'll also send it upstream separately.

(This is the awkified, Makefileified version of my little shell script

```
for test in $(awk -f testsuite/summarize.awk < testsuite/_log  | perl -ne 'print if /List of failed tests/../^$/ and s/^\s+(tests\/\S+) with \d.+/$1/'); do make -C testsuite promote TEST="$test"; done
```
)

### DIFF
--- a/Makefile.common-jst
+++ b/Makefile.common-jst
@@ -295,6 +295,14 @@ test: install_for_test
                fi \
              fi)
 
+promote-failed:
+	(export OCAMLSRCDIR=$$(pwd)/_runtest; \
+	 export CAML_LD_LIBRARY_PATH=$$(pwd)/_runtest/lib/ocaml/stublibs; \
+	 if $$(which gfortran > /dev/null 2>&1); then \
+	   export LIBRARY_PATH=$$(dirname $$(gfortran -print-file-name=libgfortran.a)); \
+	 fi; \
+	 cd _runtest/testsuite && make promote-failed)
+
 runtest-upstream: test
 
 test-one: install_for_test
--- a/testsuite/Makefile
+++ b/testsuite/Makefile
@@ -145,6 +145,7 @@ default:
 	@echo "  promote DIR=p   promote the output for the tests located in path p"
 	@echo "  promote LIST=f  promote the output for the tests listed in f (one \
 	per line)"
+	@echo "  promote-failed  promote all the tests that failed in the previous run"
 	@echo "  lib             build library modules"
 	@echo "  tools           build test tools"
 	@echo "  clean           delete generated files"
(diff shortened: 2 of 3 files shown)


riaqn:
Thanks! I like that we got rid of the parsing with `perl` and just ask `awk` to print what we need. 

Would you also care to add those to `Makefile.common-jst`? I had some issues with it because of the paths. Also it cleans up `_runtest` across runs, so I had a target `test-and-promote-failed` in one go, but for some reasons it didn't work.

antalsz:
Turns out the cleanup is only done at the *start* of `make test`, so I added a separate `make -f Makefile.jst promote-failed` target you can use in the same way.  Feel free to merge if this looks good to you!

Separately, I feel like `Makefile.common-jst` could be cleaned up.  Maybe some sort of abstraction over

```
(export OCAMLSRCDIR=$$(pwd)/_runtest; \
 export CAML_LD_LIBRARY_PATH=$$(pwd)/_runtest/lib/ocaml/stublibs; \
if $$(which gfortran > /dev/null 2>&1); then \
   export LIBRARY_PATH=$$(dirname $$(gfortran -print-file-name=libgfortran.a)); \
 fi; \
 ...)
```

and more use of `make -C` instead of `cd`?  But that's for a future PR, I'm not Makefile-hacking right now (or ever, if I can avoid it :stuck_out_tongue:).

Makefile.common-jst
```diff
@@ -295,6 +295,14 @@ test: install_for_test
                fi \
              fi)
 
+promote-failed:
```
goldfirere, line 298: I would expect this to need the `install_for_test` dependency. But that's just cargo-culted from something @stedolan did, so I might be wrong.

riaqn (approved):
Thanks - it looks good to me.

goldfirere (commented):
I can't verify the awk code, but I think that's fine. If you're happy with it, then any bugs will be glaringly obvious when they happen (and will show up in CI).

Not clicking Approve because of the `install_for_test` thing, but if you fix that, feel free to merge.

Makefile.common-jst
riaqn: (oops I have already merged it)

Yes I would want that as dependency as well; but `install_for_test` always create an empty `_runtest` dir which means `_runtest/_log` won't exist.

antalsz: Right -- this sort of "implicitly" depends on `test`, but the idea is it's run afterwards with human intervention.  `install_for_test` sets things up afresh and blows too much away.

PR #145, opened 2023-03-10
Revert "Remove `Lev_module_definition` from lambda"

ncik-roberts:
Reverts ocaml-flambda/ocaml-jst#135 due to a bad interaction with inlining. See discussion in https://github.com/ocaml/ocaml/pull/12092#issuecomment-1463977121; we'll want to reopen this later with whatever approach we decide on there.

### DIFF
--- a/bytecomp/bytegen.ml
+++ b/bytecomp/bytegen.ml
@@ -1014,6 +1014,8 @@ let rec comp_expr env exp sz cont =
             let cont1 = add_event ev cont in
             comp_expr env lam sz cont1
           end
+      | Lev_module_definition _ ->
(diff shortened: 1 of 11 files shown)


mshinwell (approved):

lthls:
I don't remember where we are on `ocaml-jst`, but on `flambda-backend` at least the fix for local functions has already been merged.
I suspect that it is also the case here, which would mean that #135 should be fine. (I'll let @mshinwell decide though.)

ncik-roberts:
Mark and I agree with your characterization---`ocaml-jst` has your fix. We don't need to revert this PR internally.

PR #146, opened 2023-03-10
Remove arity-interrupting elaboration of module patterns

ncik-roberts:
Remove the elaboration of module patterns, where typechecking would insert `let-module` expressions for each unpack. For example, this:

```ocaml
let f (module M : S) y = M.x + y
```

used to be elaborated to this during typechecking:

```ocaml
let f (m : (module S)) =
  let module M = (val m) in
  fun y -> M.x + y
```
The same went for match expressions (including `when` guards!) and let expressions (including the RHS when the let binding was recursive).

Now, we treat the existing `Tpat_var` construct in typedtree as unpacking a module pattern (as long as `pat_extra = Tpat_unpack`). To typecheck in the new scheme, we:

  * Treat module patterns as introducing a module identifier at an elevated scope
  * Take care to check that the module's types don't escape via the bound variable types or the overall expression types

The mechanism we use for this is similar to `gadt_equations_level`, in that we wrap pattern typechecking its own level and store the elevated scope in a global ref that pattern typechecking refers to. (In particular, we repurpose the `allow_modules` bool to an `int option`, where the int payload communicates the scope to enter the module identifier at.) Like `gadt_equations_level`, we only elevate the level if the pattern contains the relevant kind of subpattern (here, a module unpack).

We want to gain confidence not only that this PR avoids scope escape, but that future constructs involving patterns avoid scope escape. Previously, the way this was accomplished was via the `type_unpacks` function, which typechecked a body expression within the scope of a list of unpacked modules, and confirmed the modules' types didn't escape via the body's type. In this world, it was easy to see that scope escape was checked anywhere a module was used. Now, the responsibility is more diffuse. But we can still state the matter simply. The responsibility to check scope escape falls to whoever creates the scope such that pattern typechecking sets `allow_modules = Modules_allowed { scope }`. You can see this responsibility in action in both the `type_cases` and `type_let` functions. This is summarized in a comment by the `module_patterns_restriction` type.

`type_unpacks` now doesn't do much at all. I clean it up in a stacked PR (https://github.com/ncik-roberts/ocaml-jst/pull/1).

I add a few tests that did not always exhibit the desired behavior over the course of my feature.


typing/typecore.ml
```diff
@@ -6240,25 +6323,25 @@ and type_unpacks ?(in_function : (Location.t * type_expr * bool) option)
                    (mkloc (Longident.Lident unpack.tu_name.txt)
                       unpack.tu_name.loc)))
         in
-        Mtype.lower_nongen (get_level ty) modl.mod_type;
+        (* This call to [lower_nongen] mirrors the corresponding call in typing
+           [Ppx_letmodule]. It's not necessary here, as a first class module
```
goldfirere, line 6327:
```suggestion
           [Pexp_letmodule]. It's not necessary here, as a first class module
```

goldfirere, line 6327: So should the line be removed? The comment suggests "yes".

typing/typecore.ml
```diff
@@ -6223,14 +6303,17 @@ and type_statement ?explanation env sexp =
     exp
   end
 
-and type_unpacks ?(in_function : (Location.t * type_expr * bool) option)
+(* Type the body within an environment extended with the unpacked modules are
+   added, returning both the typechecked body and the extended environment. *)
+and type_unpacks
```
goldfirere, line 6308: It looks to me that this function can be simplified, by separating out the add-module-vars-to-env functionality to a `add_module_variables` function analogous to `add_pattern_variables`. Just like `add_pattern_variables`, the new `add_module_variables` would be called toward the end of `type_pattern_list`. And it would have to be called in `type_cases`, just after the current call to `add_pattern_variables`. The refactoring seems good, because it makes module variables more like normal variables (uniformity, yay!) -- and it avoids the second call to `type_unpacks'` in `type_cases`. (Currently, the two `type_unpacks'` calls both do the work of adding module stuff to the env't, but I think we can just do this once.) Ditto the calls to `type_unpacks` in `type_let`.

Do you think this would work? If so, then we could kill off `type_unpacks` entirely, just replacing it with `type_expect`.

typing/typecore.ml
```diff
@@ -6429,11 +6510,11 @@ and type_cases
           | None -> None
           | Some scond ->
               Some
-                (type_unpacks ext_env mode_local unpacks scond
+                (type_unpacks' ext_env mode_local unpacks scond
```
goldfirere, line 6513: Regardless of the refactoring above, I favor getting rid of `type_unpacks'`. It's just more transparent to readers to have the call to `snd` be here. Otherwise, I don't know what processing happens in `type_unpacks'` separately from `type_unpacks`.

typing/typecore.ml
```diff
@@ -6483,6 +6564,11 @@ and type_cases
   else
     (* Check for unused cases, do not delay because of gadts *)
     unused_check false;
+  if may_contain_modules then begin
+    end_def ();
+    (* Ensure that unpacked modules' types do not escape *)
+    unify_exp_types loc env (instance ty_res) (newvar ());
```
goldfirere, line 6570: This certainly looks reasonable, but I'm annoyed we have to do it. I think I want to get the calls to `instance ty_res` elsewhere to use the outer level instead of the ambient one. Let's perhaps talk about this.

goldfirere (commented):
This looks like it works well!

But I believe I found a few opportunities for simplification -- see line comments.

typing/typecore.ml
ncik-roberts: I agree, and I get rid of `type_unpacks` altogether in a stacked PR: https://github.com/ncik-roberts/ocaml-jst/pull/1

It's possible that this stacked PR should have just been included as part of this patch---I will think about doing so for the upstream PR.

typing/typecore.ml
ncik-roberts: As we discussed, I did this in https://github.com/ncik-roberts/ocaml-jst/pull/1, which I could have done more to get on your radar (sorry!).

typing/typecore.ml
```diff
@@ -734,21 +749,28 @@ let enter_variable ?(is_module=false) ?(is_as_variable=false) loc name mode ty
   if List.exists (fun {pv_id; _} -> Ident.name pv_id = name.txt)
       !pattern_variables
   then raise(Error(loc, Env.empty, Multiply_bound_variable name.txt));
-  let id = Ident.create_local name.txt in
+  let id =
+    if is_module then begin
+      (* Note: unpack patterns enter a variable of the same name *)
```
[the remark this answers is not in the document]
ncik-roberts: The comment is transplanted. It could benefit from a reword --- how's this? This is how I would answer your questions based on the new wording:
  * the module variable/value variable have the same name as each other
  * yes, this is different than non-unpack patterns, where no module declaration is entered

typing/typecore.ml
```diff
@@ -3665,6 +3698,13 @@ let may_contain_gadts p =
    | _ -> false)
   p
 
+let may_contain_modules p =
```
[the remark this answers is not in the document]
ncik-roberts: I like this suggestion, but I took it a bit further: I moved the main comment down here and replaced the other comment with a reference to the main one. WDYT?

typing/typecore.ml
ncik-roberts: The thing I was trying to suggest with this comment was: well, if one day module types *can* contain variables, then it might be better to consistently error here (as is accomplished by setting the level to 0) rather than accidentally introduce a soundness bug along the lines of the linked issue.

I also find that it's more natural to comment a line of code rather than the absence of a line of code.

However I did feel conflicted here. Let's talk about this when we talk about https://github.com/ocaml-flambda/ocaml-jst/pull/146#discussion_r1135584250.

typing/typecore.ml
```diff
@@ -3907,20 +3960,47 @@ and type_expect_
...
+
+         Checking unification within an environment extended with the module
+         bindings allows us to correctly accept more programs. This environment
+         allows unification to identify more cases where a type introduced by
+         the module is equal to a type introduced at an outer scope.
+      *)
+      if may_contain_modules then begin
+        end_def ();
+        unify_exp extended_env body (newvar ());
+        check_scope_escape_let_bound_idents extended_env pat_exp_list
```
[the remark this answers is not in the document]
ncik-roberts: I agree. Updated.

typing/typecore.ml
ncik-roberts: Typo is fixed.

typing/typecore.ml
```diff
@@ -3728,6 +3768,19 @@ let unify_exp env exp expected_ty =
   with Error(loc, env, Expr_type_clash(err, tfc, None)) ->
     raise (Error(loc, env, Expr_type_clash(err, tfc, Some exp.exp_desc)))
 
+(* Ensure that all bound idents don't leak a rigid type variable from a higher
```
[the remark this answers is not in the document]
ncik-roberts: Accepted the suggestion. My use of the word "rigid" was from one of Leo's explanations of type variables introduced by opening an existential package, though I may not have faithfully reproduced the terminology (or may have used a word that OCaml people don't normally use to refer to this). As a proposed resolution: I removed the word "rigid" as "type variable from an inner scope" seems unambiguous.

typing/typecore.ml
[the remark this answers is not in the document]
goldfirere: Looks great.

typing/typeclass.ml
```diff
@@ -1190,7 +1190,7 @@ and class_expr_aux cl_num val_env met_env virt self_scope scl =
       in
       let partial =
         let dummy = type_exp val_env (Ast_helper.Exp.unreachable ()) in
-        Typecore.check_partial val_env pat.pat_type pat.pat_loc
+        Typecore.check_partial Modules_rejected val_env pat.pat_type pat.pat_loc
```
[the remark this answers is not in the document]
ncik-roberts: The earlier call to `Typecore.type_class_arg_pattern` is what first typechecks the pattern (and this function does set `Modules_rejected`). This `check_partial` call is just for checking pattern exhaustiveness (I think?). It would probably be fine for `check_partial` to just always allow module patterns, but then you have to choose a scope to enter them at, which feels weird.

typing/typecore.ml
ncik-roberts: As discussed, we rolled this in together with the level management for gadts

typing/typecore.ml
ncik-roberts: I removed the line according to our conversation with Leo. I attempted a reword of the comment---I wanted to include an explanation beyond just "this call isn't necessary here", but I fear that my attempt isn't clarifying (and is likely wrong in some ways). Do you have a suggestion?

typing/typecore.ml
```diff
@@ -6334,6 +6334,12 @@ and type_unpacks
   let extended_env =
     List.fold_left (fun env unpack ->
       Typetexp.TyVarEnv.with_local_scope begin fun () ->
+        (* This code is parallel to the typing of Pexp_letmodule. However we
+           omit the call to [Mtype.lower_nongen] as it's not necessary here.
+           Here, we're typing an unpack which is required to have a known
+           module type; this cannot introduce new type variables which must
+           be lowered or generalized as in the case of Pexp_letmodule.
```
goldfirere, line 6341:
```suggestion
        (* This code is parallel to the typing of Pexp_letmodule. However we
           omit the call to [Mtype.lower_nongen] as it's not necessary here.
           For Pexp_letmodule, the call to [type_module] is done in a raised
           level and so needs to be modified to have the correct, outer level.
           Here, on the other hand, we're calling [type_module] outside the
           raised level, so there's no extra step to take.
```

typing/typecore.ml
```diff
@@ -690,11 +701,15 @@ type pattern_variable =
   }
 
 type module_variable =
-  string loc * Location.t
+  {
```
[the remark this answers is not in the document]
ncik-roberts: The extra field is `tu_uid : Uid.t`. I thought it would be a bit odd to create two independent `Uid.t`s for a module pattern that occurs on both sides of an or-pattern, and then drop the latter one, so I preferred not creating the `Uid.t` until after the or-pattern is processed and we're ready to add the module variable to the environment. On reflection, this doesn't seem so bad (or at least is outweighed by the benefit of reducing the number of concepts) so I'll take the suggestion.

goldfirere (approved):

PR #147, opened 2023-03-13
Annotate layout variables at the ends of layout errors

antalsz:
This improves how some layout errors are printed, so you no longer have the confusing 'a should be an instance of type 'b.

This also does some refactoring of where things live for dependency and clarity reasons!

This is a move of ccasin#5 to the new branch/repo


goldfirere:
Does this need review? Or did the review on ccasin#5 cover it?

antalsz:
No need for re-review, the only conflicts were when module qualifications had changed on the same line as something else

goldfirere:
CI is failing. Is that expected?

ccasin:
I didn't stop to think about it because it was definitely expected when this was in my fork.  Maybe it should be passing here?  The error is about -alloc-check, which is a newish feature, so at a glance I'd guess that the CI expects a recent version of ocaml-jst, and this branch is pretty old.

I have a rebased version nearly ready to go - I just need to add Antal's errors stuff to it and clean one thing up.  Why don't we wait until I've updated the branch to that to further investigate the ci (probably a day or two).

PR #149, opened 2023-03-14
Functions with no clauses aren't local-returning

ccasin:
As we observed, such functions can be constructed by a ppx.

It's kind of a pain to write a test case for this, so I haven't, but I could if it seems necessary.


riaqn (approved):
Thanks! Looks good to me.

PR #152, opened 2023-03-17
A better place to generate the `ocamlc` you use in a debug session

ncik-roberts:
`make -f Makefile.jst install` now also puts `ocamlc` at the root of the repo. It's recommended to run `ocamldebug` on it rather than the one in `$prefix/bin`.

Why's that? The `ocamldebug` session inherits the pwd of the bytecode file you pass as argument. In particular, breakpoints bring you to the proper file if you use the root `ocamlc`, and would bring you to the copied file in the `_build` directory if you used the `$prefix/bin` `ocamlc.byte`. (Well, actually, emacs would prompt you and bring you to the proper file, but it was a disruptive workflow.)


antalsz:
LGTM, although I don't use the `Makefile.jst` target myself

PR #153, opened 2023-03-22
Include the modes of values in debugging information

vouillon:
I'm considering using this information in Js_of_ocaml to generate better code when supporting effect handlers.
But it makes a lot of sense to include it in `Env.summary` in any case.


riaqn (approved):
Looks good to me. Thanks!

PR #154, opened 2023-03-24
Merge flambda-backend 2023-03-24

mshinwell:
(work by @ccasin )


mshinwell:
The testsuite failures only seem to be happening on the i386 check, which isn't relevant because (a) it has stack allocation disabled, yet our assumption is for that to be enabled everywhere now; and (b) it's a 32-bit platform.  I think it's (a) causing the problem in this case.  I will make a follow-up PR to remove this check.

PR #155, opened 2023-03-24
Remove i386 CI check

mshinwell:
See https://github.com/ocaml-flambda/ocaml-jst/pull/154#issuecomment-1482678147


goldfirere (approved):

goldfirere:
Please merge after CI passes.

Ekdohibs:
Note that the CI failures were only for the compilation of primitives, which now have layout information refined by their type. Simply updating the corresponding files by promoting the test results would be enough here.

PR #156, opened 2023-03-24
Local immutable arrays

antalsz:
Comprehensions are coming in a separate PR


stdlib/iarray.ml
```diff
@@ -169,57 +392,137 @@ let fold_right f a x =
...
   !r
 
+(* CR aspectorzabusky: Why do I need this?  Shouldn't mode-crossing handle doing
+   this? *)
+let[@inline always] globalize_bool : local_ bool -> bool = fun b -> b
+
 let exists p a =
   let n = length a in
-  let rec loop i =
+  let rec loop i = local_
```
goldfirere, line 401: I don't think you really need the `local_` here, or in the `local` version. Ditto similar functions below.

stdlib/iarray.ml
```diff
@@ -230,6 +533,18 @@ let find_opt p a =
       if p x then Some x
       else loop (succ i)
   in
+  (* CR aspectorzabusky: Does this really have to be [@nontail]? *)
+  loop 0 [@nontail]
```
goldfirere, line 537: Yes, I think so. `loop` closes over the local `p`.

stdlib/iarray.ml
```diff
@@ -241,12 +556,46 @@ let find_map f a =
       | None -> loop (succ i)
       | Some _ as r -> r
   in
+  (* CR aspectorzabusky: Does this really have to be [@nontail]? *)
```
goldfirere, line 559: Yep.

stdlib/iarray.ml
```diff
@@ -241,12 +556,46 @@ let find_map f a =
...
+let find_map_local_input f a =
+  let n = length a in
+  let rec loop i =
+    if i = n then None
+    else
+      match f (unsafe_get a i) with
+      | None -> loop (succ i)
+      | Some _ as r -> r
+  in
+  (* CR aspectorzabusky: Does this really have to be [@nontail]? *)
```
goldfirere, line 582: Yep.

stdlib/iarray.mli
```diff
@@ -98,111 +118,319 @@ val of_array : 'a array -> 'a iarray
...
+val mapi_local_output :
+  local_ (int -> 'a -> local_ 'b) -> 'a iarray -> local_ 'b iarray
+(** The locally-allocating but global-input version of [mapi]. *)
+
+val fold_left : local_ ('a -> 'b -> 'a) -> 'a -> 'b iarray -> 'a
 (** [fold_left f init a] computes
    [f (... (f (f init a.:(0)) a.:(1)) ...) a.:(n-1)],
    where [n] is the length of the immutable array [a]. *)
 
+(* MISSING: Requires [let mutable]
```
goldfirere, line 176: Really? I don't see why.

stdlib/iarrayLabels.mli
```diff
@@ -104,111 +124,307 @@ val of_array : 'a array -> 'a iarray
...
+  f:local_ ('a -> local_ 'b -> unit) -> 'a iarray -> local_ 'b iarray -> unit
+(** The second-biased partly-locally-scoped version of [iter2]. *)
+
+val map2 : f:local_ ('a -> 'b -> 'c) -> 'a iarray -> 'b iarray -> 'c iarray
 (** [map2 ~f a b] applies function [f] to all the elements of [a]
    and [b], and builds an immutable array with the results returned by [f]:
    [[| f a.:(0) b.:(0); ...; f a.:(length a - 1) b.:(length b - 1)|]].
    @raise Invalid_argument if the immutable arrays are not the same size. *)
 
+val map2_local :
```
goldfirere, line 289: iarray.mli has more varieties of these functions

stdlib/iarrayLabels.mli
```diff
@@ -38,10 +38,11 @@ open! Stdlib
 type +'a t = 'a iarray
 (** An alias for the type of immutable arrays. *)
 
-external length : 'a iarray -> int = "%array_length"
+external length : local_ 'a iarray -> int = "%array_length"
```
goldfirere, line 41: Some of the comments in this file have diverged from iarray.mli. They should probably be lined up.

typing/typecore.ml
```diff
@@ -6879,18 +6878,20 @@ and type_generic_array
...
       ~ty_expected
       ~explanation
-      ~type_
       ~mutability
       ~attributes
       sargl
   =
+  let type_, base_argument_mode = match mutability with
+    | Mutable -> Predef.type_array, mode_default Value_mode.global
+    | Immutable -> Predef.type_iarray, mode_subcomponent expected_mode
```
goldfirere, line 6887: @riaqn: Does the `mode_subcomponent` here look right to you? Feel free to message me or Antal to explain more context here.

lambda/lambda.mli
```diff
@@ -195,6 +195,10 @@ type primitive =
                         one; O(1) *)
   | Parray_of_iarray (* Unsafely reinterpret an immutable array as a mutable
                         one; O(1) *)
+  | Pempty_iarray (* The equivalent of [[::]], but available without any
```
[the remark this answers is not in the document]
antalsz: 1. I use it because we need to create empty arrays in `stdlib/iarray.ml`.
2. I don't know how to get the build system to do that; I'd be perfectly happy to.
3. Before this, I cast `[||]` to `[::]` in `stdlib/iarray.ml`; now I have to use it directly, because I have to be able to put it on the local stack.
4. We can't use the expanded forms because the parser can't tell the difference and is still unhappy.

middle_end/closure/closure.ml
```diff
@@ -1326,6 +1326,9 @@ let rec close ({ backend; fenv; cenv ; mutable_vars; kinds; catch_env } as env)
            Pobj_magic),
           [arg], _loc) ->
       close env arg
+  | Lprim(Pempty_iarray, [arg], _loc) ->
```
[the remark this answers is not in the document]
antalsz: You can't have 0-ary primitives, IIRC

stdlib/iarray.ml
```diff
@@ -56,86 +56,310 @@ type +'a t = 'a iarray
...
+external make_mutable_local : int -> local_ 'a -> local_ 'a array =
+  "caml_make_local_vect"
+external unsafe_of_local_array : local_ 'a array -> local_ 'a iarray =
+  "%array_to_iarray"
+external unsafe_sub_of_array_local :
+  local_ 'a array -> int -> int -> local_ 'a iarray = "caml_array_sub_local"
+external unsafe_set_local : local_ 'a array -> int -> local_ 'a -> unit =
+  "%array_unsafe_set"
+
+(* We can't use immutable array literals here, since we don't want to require
```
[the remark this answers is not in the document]
antalsz: We've gone with a command-line based syntax rather than a Haskell-style pragma based syntax, so it would be something in the makefile that I don't know how to do.  See above re. the explicit syntax.

stdlib/iarray.ml
```diff
@@ -56,86 +56,310 @@ type +'a t = 'a iarray
...
+   the stdlib to be compiled with extensions *)
+external empty_iarray : unit -> 'a iarray =
+  "%empty_iarray"
```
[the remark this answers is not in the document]
antalsz: Calling a C function is more expensive, and we can't convert an empty array anymore (we used to) because we can't put the empty array on the local stack.

stdlib/iarray.ml
```diff
@@ -56,86 +56,310 @@ type +'a t = 'a iarray
...
+
+(* CR aspectorzabusky: Really trusting the inliner here, it has to inline
```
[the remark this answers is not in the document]
antalsz: For performance, I should clarify that

stdlib/iarray.ml
```diff
@@ -56,86 +56,310 @@ type +'a t = 'a iarray
...
+external unsafe_of_local_array : local_ 'a array -> local_ 'a iarray =
+  "%array_to_iarray"
+external unsafe_sub_of_array_local :
+  local_ 'a array -> int -> int -> local_ 'a iarray = "caml_array_sub_local"
+external unsafe_set_local : local_ 'a array -> int -> local_ 'a -> unit =
+  "%array_unsafe_set"
+
+(* We can't use immutable array literals here, since we don't want to require
+   the stdlib to be compiled with extensions *)
+external empty_iarray : unit -> 'a iarray =
```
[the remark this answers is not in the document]
antalsz: AIUI, that's a rule for externals

stdlib/iarray.ml
```diff
@@ -56,86 +56,310 @@ type +'a t = 'a iarray
...
+      let x = f i in
+      if i = l - 1 then
+        make_mutable_local l x
+      else begin
+        let res = go (i+1) in
+        unsafe_set_local res i x;
+        res
+      end
+    end in
+    unsafe_sub_of_array_local (go 0) 0 l
```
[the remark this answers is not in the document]
antalsz: That's the step that puts on on the correct stack, right?  Or wait, maybe that isn't necessary...

stdlib/iarray.ml
```diff
@@ -169,57 +392,137 @@ let fold_right f a x =
   done;
   !r
 
+(* CR aspectorzabusky: Why do I need this?  Shouldn't mode-crossing handle doing
```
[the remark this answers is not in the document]
antalsz: I agree but I did at the time.  I'll try again after a rebase.

stdlib/iarray.ml
antalsz: I'm fairly sure I only added that as requested but I can double-check.

stdlib/iarray.mli
antalsz: If you look at the implementation in `array.ml`, you'll see that it uses a locally-scoped `ref`; we can't put local values in there, which is exactly what `let mutable` solves.

lambda/lambda.mli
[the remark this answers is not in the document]
goldfirere: These are all good answers. But I'd still say all of the fuss in this patch to make a new primitive isn't worth it. I see the following possible ways forward:

- The manual documents `ocaml_compiler_internal_params` (see just before https://v2.ocaml.org/manual/comp.html#s:modules-file-system). That looks like it does what you want to enable an extension for an individual file. It's processed in `driver/compenv.ml` in the `apply_config_file` function. It looks like you'd have to reverse-engineer the file format a bit, but it doesn't look too hard.
- I think it's actually OK to have an extension on for the entire stdlib, if the above suggestion doesn't pan out. I'm not exactly sure where to put the extension, but `stdlib/Makefile` has `COMPFLAGS` which looks like a very good starting point. Not sure where the dune equivalent is, or if it's necessary.
- I seem to recall some discussion of a way of enabling an extension in a file. But I can't find evidence of this. Perhaps ask around on Slack.
- Failing all that, I think adding a new C function that just creates an empty array is still much less invasive than a whole new primitive.
- And, actually, I don't see a great reason not to just make an empty normal array and cast it to an immutable one. I seem to recall that empty arrays are shared on the heap, so it wouldn't do any allocation anyway. (Double-check this.)

middle_end/closure/closure.ml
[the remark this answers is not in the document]
goldfirere: `Pgetglobal` appears to be a 0-ary primitive. I certainly don't know the details, but it looks possible.

Even if we need an argument.... what is it? There should be a comment somewhere.

stdlib/iarray.ml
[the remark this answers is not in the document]
goldfirere: Oh. That's true about the C function. Well, cross that off the list of ideas above. But there are still many other ideas there to pursue.

stdlib/iarray.ml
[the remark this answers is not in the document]
goldfirere: I *think* it's already on the stack. If it weren't, your tests would show allocations, I think.

stdlib/iarray.mli
goldfirere: Well, yes, but it looks possible to avoid the `ref`:

```
let fold_left_local :
  local_ (local_ 'a -> local_ 'b -> local_ 'a) ->
  local_ 'a ->
  local_ 'b iarray ->
  local_ 'a =
  fun f z arr -> local_
  let len = length arr in
  let rec go acc i = local_
    if i = len
    then acc
    else go (f acc (unsafe_get arr i)) (i+1)
  in
  go z 0
```

This is still not that great, because it permanently consumes O(n) stack space (well, until the outer call's region ends). But I don't see a way to avoid the stack consumption, because the final value must go *somewhere*. By induction, all previous values must go *somewhere*. And each one can't be gotten rid of until the next one is computed. So unless the function gets some kind of `copy:('a -> 'a)` argument, I don't see a way to avoid the stack usage. Maybe someone else can.

Note that the `let mutable` solution is no better in this regard: each intermediate `'a` would end up sitting on the stack until the outer caller's region is popped.

Do add a comment in the mli file warning users that all their intermediate values will be put on the stack. In the end, they probably don't want this function, but it's not for us to tell users what they do and don't want.

lambda/lambda.mli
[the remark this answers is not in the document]
lpw25: Do I understand correctly that you are trying to make a local empty immutable array? I think you can continue to just cast a `[||]` because that is never an actual allocation at runtime.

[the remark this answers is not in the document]
antalsz: Not only was this sensible advice, but it turns out we didn't actually use any local `[::]`s anymore :man_facepalming:.  `%empty_iarray` is gone again.

antalsz:
This is all wrapped up except for rebasing as of 3be589b8342619e9b17ce43348a001f52845f9e2

goldfirere:
As discussed just as I was leaving yesterday, while I would still like a quick comment from e.g. @riaqn, I think this is OK to merge.... except for the CI failures above.

typing/typecore.ml
riaqn: I'm terribly sorry about the delay, yes it looks correct to me. A value being `regional` means it's anywhere except the current region. If an array is `regional`, we require its content to be `global`, because requiring `regional` is too vague - the content might be in a region "inner" than the array and get deleted, in which case the array would be pointing to an invalid address. @lpw25 will correct me if I'm wrong.

antalsz:
This is ready for re-review now that I've made the array get/set primitives mode-aware; we should also get somebody who understands backend stuff (and maybe modes?) to review those parts of it given those changes

mshinwell:
I'm reading the middle/backend parts, sorry for the delay

asmcomp/cmm_helpers.mli
```diff
@@ -555,9 +559,12 @@ val bigstring_load :
 
 (** Arrays *)
 
-(** Array access. Args: array, index *)
-val arrayref_unsafe : Lambda.array_kind -> binary_primitive
-val arrayref_safe : Lambda.array_kind -> binary_primitive
+(** Array access.  The [alloc_mode] is used only if the [array_kind] is
+    [Pfloatarray]; the resulting [float] will then be allocated at the specified
+    mode.
```
mshinwell, line 564: Again, change the type so the condition is statically checked and doesn't need a comment.

asmcomp/cmm_helpers.ml
```diff
@@ -790,23 +790,29 @@ let int_array_ref arr ofs dbg =
...
-let addr_array_set arr ofs newval dbg =
+let addr_array_set_heap arr ofs newval dbg =
   Cop(Cextcall("caml_modify", typ_void, [], false),
       [array_indexing log2_size_addr arr ofs dbg; newval], dbg)
+
+let addr_array_set_local arr ofs newval dbg =
+  Cop(Cextcall("caml_modify_local", typ_void, [], false),
+      [arr; untag_int ofs dbg; newval], dbg)
+
+let addr_array_set (mode : Lambda.modify_mode) =
```
mshinwell, line 804: This would be clearer if it were eta expanded.

runtime/array.c
```diff
@@ -181,6 +270,22 @@ CAMLprim value caml_array_unsafe_set(value array, value index, value newval)
   return caml_array_unsafe_set_addr(array, index, newval);
 }
 
+/* [ local_ 'a array -> int -> local_ 'a -> unit ]
+
+   Must be used carefully, as it can violate the "no forward pointers"
+   restriction on the local stack. */
```
mshinwell, line 276: Should the setting functions above with the bounds checks also have this comment?

mshinwell:
Once comments have been addressed, please turn this into an flambda-backend PR: we should do the flambda2 implementation before this is merged.

lambda/transl_array_comprehension.ml
```diff
@@ -756,7 +756,9 @@ let body
   let open Let_binding in
   let set_element_raw elt =
     (* array.(index) <- elt *)
-    Lprim(Parraysetu array_kind, [array.var; index.var; elt], loc)
+    Lprim(Parraysetu (Lambda.modify_heap, array_kind),
```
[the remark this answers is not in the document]
antalsz: Correct; changing that is hypothetically on my stack but considered low priority

lambda/lambda.mli
```diff
@@ -127,10 +127,18 @@ type primitive =
       The arguments of [Pduparray] give the kind and mutability of the
       array being *produced* by the duplication. *)
   | Parraylength of array_kind
-  | Parrayrefu of array_kind
-  | Parraysetu of array_kind
-  | Parrayrefs of array_kind
-  | Parraysets of array_kind
+  | Parrayrefu of alloc_mode * array_kind
+  (** The [alloc_mode] only matters for flat float arrays; accessing them must
+      allocate the resulting [float]. *)
```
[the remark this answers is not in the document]
antalsz: Thanks, this caught a few things; I made this change in 3c4b26d40.  One specific question: [in translprim.ml, lines 950--953](https://github.com/ocaml-flambda/ocaml-jst/blob/3c4b26d40aa2942805643a2171ce261710cd6516/lambda/translprim.ml#L950-L953), there's code that asserts that unsafely setting an element of a possibly-float array could raise an exception or allocate; this doesn't seem right to me.  What am I missing?

lambda/translprim.ml
```diff
@@ -225,18 +225,54 @@ let lookup_primitive loc poly pos p =
...
+       in
+       Primitive ((Parraysetu (obj_mode, gen_array_kind)), 3)
     | "%floatarray_length" -> Primitive ((Parraylength Pfloatarray), 1)
-    | "%floatarray_safe_get" -> Primitive ((Parrayrefs Pfloatarray), 2)
-    | "%floatarray_safe_set" -> Primitive ((Parraysets Pfloatarray), 3)
-    | "%floatarray_unsafe_get" -> Primitive ((Parrayrefu Pfloatarray), 2)
-    | "%floatarray_unsafe_set" -> Primitive ((Parraysetu Pfloatarray), 3)
+    | "%floatarray_safe_get" -> Primitive ((Parrayrefs (mode, Pfloatarray)), 2)
+    | "%floatarray_safe_set" ->
+       (* The mode doesn't actually matter for floatarrays, it turns out *)
```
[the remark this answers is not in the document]
antalsz: Done in 3c4b26d40aa2942805643a2171ce261710cd6516; see above

lambda/lambda.mli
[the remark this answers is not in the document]
antalsz: Missed some; further improvements in 2f3811e0c669a59bd1f44390c63f98d48abcf88d

asmcomp/cmm_helpers.mli
antalsz: Done in 2f3811e0c669a59bd1f44390c63f98d48abcf88d; see above

middle_end/clambda_primitives.mli
```diff
@@ -77,10 +79,18 @@ type primitive =
       The arguments of [Pduparray] give the kind and mutability of the
       array being *produced* by the duplication. *)
   | Parraylength of array_kind
-  | Parrayrefu of array_kind
-  | Parraysetu of array_kind
-  | Parrayrefs of array_kind
-  | Parraysets of array_kind
+  | Parrayrefu of alloc_mode * array_kind
```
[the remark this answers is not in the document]
antalsz: Done in 2f3811e0c669a59bd1f44390c63f98d48abcf88d; see above

antalsz:
I'm going to leave this open for now but @mshinwell, I've responded to all your review (with one open question) and have moved this over to ocaml-flambda/flambda-backend#1420

lambda/translprim.ml
```diff
@@ -902,8 +935,10 @@ let lambda_primitive_needs_event_after = function
   | Paddfloat _ | Psubfloat _ | Pmulfloat _ | Pdivfloat _ | Pstringrefs | Pbytesrefs
   | Pbox_float _ | Pbox_int _
   | Pbytessets | Pmakearray (Pgenarray, _, _) | Pduparray _
-  | Parrayrefu (_, (Pgenarray | Pfloatarray))
-  | Parraysetu (_, (Pgenarray | Pfloatarray))
+  | Parrayrefu (Pgenarray_ref _ | Pfloatarray_ref _)
+  | Parraysetu (Pgenarray_set _ | Pfloatarray_set)
+    (* CR aspectorzabusky for mshinwell: It seems to be that unsafe set should
+       never allocate or throw an exception; am I missing something? *)
```
mshinwell, line 941: Agreed, it seems odd that these cases are here.

lambda/translprim.ml
```diff
@@ -471,14 +449,68 @@ let simplify_constant_constructor = function
...
+  | (Paddrarray_ref | Pintarray_ref), Pfloatarray -> t1
+
+  (* Compute the correct GLB *)
+
+  (* Pgenarray >= _ *)
+  | (Pgenarray_ref _ as x), Pgenarray -> x
+  | Pgenarray_ref _, Pintarray -> Pintarray_ref
+  | Pgenarray_ref _, Paddrarray -> Paddrarray_ref
+  | Pgenarray_ref mode, Pfloatarray -> Pfloatarray_ref mode
+  | (Paddrarray_ref | Pintarray_ref | Pfloatarray_ref _ as x), Pgenarray -> x
```
mshinwell, line 475: To better match the code in `glb_array_type`, maybe fold the last case into the first one.

mshinwell: (same in `glb_array_set_type` below)

mshinwell (changes requested):
Just a couple of minor things.

lambda/translprim.ml
antalsz: I don't understand this comment -- in `glb_array_type`, things are separated out exactly the same way as I have it, with `float, addr | int` and `addr | int, float` cases leading things off.

lambda/translprim.ml
antalsz: Done on flambda-backend in 123629587

lambda/translprim.ml
antalsz: @mshinwell What's the verdict here?

mshinwell: OK, let's leave this then.

antalsz:
Closing in favor of ocaml-flambda/flambda-backend#1420

PR #157, opened 2023-03-24
Unboxed layout histories

antalsz:
This PR adds histories to layouts, so that we can keep track of why they changed and provide better error messages.  This patch mostly just provides the support, not the error messages.


goldfirere:
Closed in favor of #158, with a cleaner history.

PR #158, opened 2023-03-27
Unboxed layout histories

goldfirere:
This PR adds histories to layouts, so that we can keep track of why they changed and provide better error messages. This patch mostly just provides the support, not the error messages.

(This is by @antalsz and supersedes #157, but with a cleaner history for better reviewability.)


goldfirere (commented):
Timing out for now, but I wanted to post what I had.

For the upcoming release, it might be best to not print out the histories, giving us time to refine this design. (I still think the basic design is a good one, but there are some implementation wrinkles.) But figuring out how to get "missing cmi file" in the output of the errors around missing cmi files would be a great point to focus on.

typing/typedecl.ml
```diff
@@ -1398,7 +1429,7 @@ let transl_extension_constructor ~scope env type_path type_params
         let args =
           update_constructor_arguments_layouts env sext.pext_loc args layouts
         in
-        let constant = Array.for_all Layout.(equate void) layouts in
+        let constant = Array.for_all Layout.(equal void) layouts in
```
goldfirere, line 1432: Why are we sure that `layouts` has no sort variables?

typing/typedecl.ml
```diff
@@ -1567,7 +1598,7 @@ let transl_type_extension extend env loc styext =
   | None -> ()
   | Some err -> raise (Error(loc, Extension_mismatch (type_path, env, err)))
   end;
-  let ttype_params = make_params env styext.ptyext_params in
+  let ttype_params = make_params env type_path styext.ptyext_params in
```
goldfirere, line 1601: This `type_path` is potentially confusing:

```ocaml
type 'a t = ..
type ('a : immediate) t += Foo
```

The output will make it look like the problem is on the first line, not the second.... but these lines might even be in different files! This would be solved if we can store the source location of the definition.

typing/typedecl.ml
```diff
@@ -1869,7 +1900,8 @@ let transl_with_constraint id ?fixed_row_path ~sig_env ~sig_decl ~outer_env
      declaration [sdecl] in the outer environment [outer_env]. *)
   let env = outer_env in
   let loc = sdecl.ptype_loc in
-  let tparams = make_params env sdecl.ptype_params in
+  (* CR aspectorzabusky: Log that it's from a constraint? *)
```
goldfirere, line 1903: I think so -- otherwise the error message is going to be very confusing.

But actually I'm not sure that an error can occur here. I don't think we allow e.g. `S with type ('a : immediate) t = ...`. Would any layout checks on the RHS of the `=` be ascribed this reason?

typing/types.ml
```diff
@@ -418,6 +418,28 @@ let equal_tag t1 t2 =
   | Extension (path1,_), Extension (path2,_) -> Path.same path1 path2
   | (Ordinary _ | Extension _), _ -> false
 
+let equal_variant_representation r1 r2 = match r1, r2 with
```
goldfirere, line 421: Worth shortcutting with `==`? (Honest question -- not sure what best practice is.)

typing/typetexp.ml
```diff
@@ -524,7 +529,10 @@ and transl_type_aux env policy mode styp =
                    List.map (transl_type env policy Alloc_mode.Global) stl)
             in
             List.iter (fun {ctyp_type; ctyp_loc} ->
-              match constrain_type_layout env ctyp_type Layout.value with
+              match
+                constrain_type_layout ~reason:Dummy_reason_result_ignored
```
goldfirere, line 533: Actually, I've figured out what's potentially wrong with `Dummy_reason_result_ignored`: the layout can get preserved in the layout of a tyvar. So even if we check against `Ok _`, it might get printed someday. This is probably true everywhere we use `constraint_type_layout` (as opposed to `check_type_layout`).

utils/misc.ml
```diff
@@ -222,6 +222,17 @@ module Stdlib = struct
...
       | Exit -> None
+
+    let equal eq_elt l1 l2 =
+      (* Basically inlines [Array.for_all2] to avoid the [raise] *)
+      let n = Array.length l1 in
+      Int.equal n (Array.length l2) &&
+      let rec loop i =
+        Int.equal i n ||
+        (eq_elt (Array.unsafe_get l1 i) (Array.unsafe_get l2 i) &&
+         loop (succ i))
```
goldfirere, line 233: This isn't tail-recursive. :(

(Unless `||` and `&&` are. Which they might be. But the docs don't seem to guarantee it, so I wouldn't rely on it.)

goldfirere (commented):
OK. Finished.

I think the quickest way to merge this is to remove the `equate`/`equal` distinction (there are two many `equal`s, and they worry me) and stop printing out the histories. That will leave a little bit of useless code, but we'll make it useful soon enough.

typing/layouts.ml
```diff
@@ -208,82 +267,183 @@ module Layout = struct
...
+  let equate = equate_or_equal ~allow_mutation:true
+
+  let intersection ~reason l1 l2 =
     let err = Error (Violation.No_intersection (l1, l2)) in
     let equality_check is_eq l = if is_eq then Ok l else err in
     (* it's OK not to cache the result of [get], because [get] does path
        compression *)
-    match get l1, get l2 with
-    | Const Any, _ -> Ok l2
+    Result.map (add_reason reason) @@ match get l1, get l2 with
```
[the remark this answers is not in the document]
antalsz: I don't understand this comment – what are you getting at?

testsuite/tests/typing-layouts/basics.ml
```diff
@@ -55,7 +55,8 @@ Line 2, characters 10-29:
 2 |   val f : void_unboxed_record -> int
               ^^^^^^^^^^^^^^^^^^^
 Error: Function argument types must have layout value.
-        void_unboxed_record has layout void, which is not a sublayout of value.
+        void_unboxed_record has layout void,
```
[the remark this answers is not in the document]
antalsz: I added line breaking in general, and `void_unboxed_record` was long enough to trigger it – we need *some* line breaking

testsuite/tests/typing-layouts/basics.ml
```diff
@@ -95,10 +99,16 @@ Line 4, characters 35-41:
 Error: The type constraints are not consistent.
        Type ('a : value) is not compatible with type t
        t has layout void, which is not a sublayout of value.
+       The latter was constrained during unification with 'a
+       and during unification with 'a
```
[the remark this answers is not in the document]
antalsz: I bet it is a different one, and we're just printing too much

utils/misc.ml
antalsz: I'm 80% sure they are, having read the compiler source, but might be worth tweaking for clarity regardless

typing/layouts.ml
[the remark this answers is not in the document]
goldfirere: Well, I thought that a new history item would be added only when the layout actually changes. But this adds a new history item unconditionally.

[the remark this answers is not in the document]
antalsz: What I didn't understand was the second part – "maybe a layout becomes `immediate` because of many different checks?"

typing/typedecl.ml
```diff
@@ -88,6 +88,14 @@ let get_unboxed_from_attributes sdecl =
   | false, true -> Some true
   | false, false -> None
 
+(* Used for layout error reporting *)
+(* CR aspectorzabusky: This feels like it must exist somewhere *)
+let parameter_name sty = match sty.ptyp_desc with
+  | Ptyp_any -> "_"
+  | Ptyp_var name -> "'" ^ name
+  | _ -> Misc.fatal_error
```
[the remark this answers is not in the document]
antalsz: Then what should we do?

typing/ctype.ml
```diff
@@ -1947,15 +1947,18 @@ let rec estimate_type_layout env ty =
    it's a valid argument to [t].  (We believe there are still loops like this
    that can occur, though, and may need a more principled solution later).
 *)
-let rec constrain_type_layout ~fixed env ty1 layout2 fuel =
+let rec constrain_type_layout ~reason ~fixed env ty1 layout2 fuel =
+  (* XXX ASZ: Are all the [~reason]s in this function the same? *)
```
goldfirere, line 1951: Our convention is that we don't merge `XXX` comments. Maybe just change this to `CR-soon` or something?

typing/types.ml
```diff
@@ -418,6 +418,28 @@ let equal_tag t1 t2 =
   | Extension (path1,_), Extension (path2,_) -> Path.same path1 path2
   | (Ordinary _ | Extension _), _ -> false
 
+let equal_variant_representation r1 r2 = r1 == r2 || match r1, r2 with
```
goldfirere, line 421: Why short-circuit here but not in `equal_record_representation`? I really don't know the value of these short-circuit checks -- but we should be principled about them, either doing them everywhere, nowhere, or informed by perf checks.

typing/ctype.ml
ccasin: My impression was our convention is not to merge `XXX`s to `main`, but it feels reasonable to merge them to the unboxed types branch.

ccasin: (And indeed if we want to merge this but also make sure this issue is resolved before merging to main, leaving a `Layouts XXX` feels the best thing to do?)

goldfirere: Yes, I agree. My impression was that this particular comment was OK to merge to `main`, which is why I suggested the change. That is, I was worried about a second merge of this code, not the merge to `unboxed-types-v1`.

goldfirere:
This was overtaken by events a long time ago.

PR #162, opened 2023-04-20
Fix bug where incorrect levels from type_package led to scope escape

ccasin:
This fixes a bug Leo and I spotted this morning.  It becomes more of a problem with layouts.

I submitted the fix simultaneously upstream - [see here](https://github.com/ocaml/ocaml/pull/12195) for the full explanation.


goldfirere:
Should this perhaps be marked as draft while the upstream patch is pending? I went to review this, but then decided not to. :)

ccasin:
Closed.  This fix is already merged (in the layouts PR), so my attempt to separate concerns by making this PR failed.  If the alternate solution proposed upstream is one day implemented, we can easily back this out and switch to that.

PR #163, opened 2023-04-21
Basic (shared) borrowing 

riaqn:
(WIP) Based on https://github.com/ocaml-flambda/ocaml-jst/pull/138
This is to support basic shared borrowing.

- [x] let binding borrowing
- [x] func application borrowing
  - [x] More tests needed  
- [x] match-binding borrowing
- [x] parser support
  - [x] make & combines tigher
- [x] fix closure over borrowing


riaqn:
We have now used the new approach of borrow-bindings (instead of shallow syntax detection). Currently only several language constructs are borrow-binders: `let`, `match` and function application. I need to go through all language constructs and ensure:
- Some other language constructs should act as borrow-binders.
- LC that doesn't bind borrowing should probably pass-through borrow-binding status to subexpressions.

Both above doesn't affect the soundness, but the completeness.

riaqn:
superseded by https://github.com/ocaml-flambda/flambda-backend/pull/1997

PR #164, opened 2023-04-24
Missing cmi errors

antalsz:
This PR adds very direct support for errors for missing `.cmi` files.  We can hopefully replace this with a history-based mechanism at some point.


typing/layouts.ml
```diff
@@ -388,23 +394,74 @@ module Layout = struct
   module Violation = struct
     open Format
 
+    let printtyp_path = ref (fun _ _ -> assert false)
+
+    let set_printtyp_path f = printtyp_path := f
```
ccasin, line 399: Not related to your patch, but this makes me realize that we should just play a similar game with `type_expr` and simplify the types of `report_with_offender`.  That function is always passed `Printtyp.type_expr` - I thought I would end up passing other things, but in the end there's no reason it needs its (pretty confusing) type.  Maybe leave a CR to consider that change, while you're here.

typing/layouts.mli
```diff
@@ -84,8 +84,10 @@ module Layout : sig
 
   (** This layout is the top of the layout lattice. All types have layout [any].
       But we cannot compile run-time manipulations of values of types with layout
-      [any]. *)
-  val any : t
+      [any].
+
+      If [missing_cmi_for] is [None], this value is guaranteed to be shared. *)
+  val any : missing_cmi_for:Path.t option -> t
```
ccasin, line 90: I'll stop short of suggesting a default argument, but maybe we should have `any_not_missing` or something - this gets pretty verbose in the code.

typing/layouts.ml
```diff
@@ -491,10 +548,12 @@ end
     open Format
 
     let internal ppf : internal -> unit = function
-      | Any         -> fprintf ppf "Any"
-      | Sort s      -> fprintf ppf "Sort %a" Sort.Debug_printers.t s
+      | Any { missing_cmi_for } ->
+          fprintf ppf "Any { missing_cmi_for = %a }"
+            (Misc.Stdlib.Option.print Path.print) missing_cmi_for
```
ccasin, line 553: A problem I have a lot is the typed tree printer is so verbose it can be hard to find the relevant bits.  What do you think about only printing this record if it's Some?  (I do not feel strongly either way)

PR #165, opened 2023-04-28
update HACKING.jst.adoc

riaqn:
add some instructions to benchmark using `typecore.ml`

### DIFF
--- a/HACKING.jst.adoc
+++ b/HACKING.jst.adoc
@@ -82,3 +82,26 @@ To add a new printer, simply add a line of the form
 
 to `tools/debug_printers.ml`, and then run `make debug_printers` in the `tools`
 directory to regenerate the printing script.
+
+## Benchmarking
+
+A small but relatively comprehensive benchmark is to run our compiler against `typing/typecore.ml`. First we install the 
+opam switch with FP (frame pointers) enabled; adjust the version number as needed:
+
+    $ opam switch create 4.14.1-fp --packages=ocaml-variants.4.14.1+options,ocaml-option-fp --repos=default
+    
+Remember to check that the newly installed switch is being used:
+
+    $ opam switch
+    #   switch     compiler                                         description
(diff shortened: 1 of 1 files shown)


HACKING.jst.adoc
```diff
@@ -82,3 +82,26 @@ To add a new printer, simply add a line of the form
...
+    ->  4.14.1-fp  ocaml-option-fp.1,ocaml-variants.4.14.1+options  4.14.1-fp
+
+Then build the compiler - the following command will build the compiler using the opam switch, then use the newly-built compiler to build itself.
+
+    $ make -f Makefile.jst compiler
+    
+We can now benchmark our compiler against `typecore.ml`. The following `_bootinstall` is built using the opam switch and has FP enabled. 
+
+    $ cd _build/main
+    $ perf stat -r 5 ../_bootinstall/bin/ocamlc.opt -strict-sequence -principal -w +a-4-9-40-41-42-44-45-48-66-70 -warn-error A -bin-annot -safe-string -strict-formats -w -67 -g -bin-annot -I .ocamlcommon.objs/byte -I ../install/runtime_stdlib/lib/ocaml_runtime_stdlib/  -intf-suffix .ml -no-alias-deps -o .ocamlcommon.objs/byte/typecore.cmo -c -impl typecore.ml
```
goldfirere, line 107: My guess is that you're running this command in `typing` and that you passed `--prefix ./install` to `./configure`. Is that right? If so, could you add these details? I pass a different directory name to `configure` and would have to modify the command line.

goldfirere (commented):
Helpful instructions!

HACKING.jst.adoc
riaqn: Actually `$(REPO)/_build/main/typecore.ml` exists for some unknown reason - so the above command runs in the repo root directory. `$(REPO)_build/install` also exists, which is (I guess) a staging area that _will be_ installed upon `make install`.

goldfirere (approved):

PR #167, opened 2023-06-18
Fix error in data-race-freedom.md

bcc32:
Global modes are less strict than local modes, for example, so each entry in a column is less strict than the one above.


bcc32:
Never mind, I misread.

PR #168, opened 2023-10-13
Resolve conflicts in configuration and dune files

stedolan:
This is enough to get `./configure` to work and to get `dune` to start and generate compiler errors (instead of choking on bad dune files).

This also changes the dune build profile for the boot compiler to `dev`, which compiles all files in opaque mode (meaning that you do not need to fix conflicts in your dependencies implementations, just their interfaces)


utils/misc.ml
```diff
@@ -375,7 +375,9 @@ let find_in_path_uncap path name =
 
 let remove_file filename =
   try
-    if Sys.is_regular_file filename
+    (* Partial revert of upstream PR 11412, to allow building with OCaml 4.14
+       (which does not have is_regular_file *)
```
lpw25, line 379: Maybe mark this comment with something grepable.

lpw25 (approved):
One small bug, otherwise looks good

PR #169, opened 2023-10-18
Fix conflicts in parsing/utils mlis (and some easy ml ones)

ccasin:


ncik-roberts (approved):
LGTM

ncik-roberts (commented):
Actually I see a few build errors:

```
$ dune build --workspace=duneconf/boot.ws _build/default/.ocamlcommon.objs/native/clflags.cmx
File "utils/clflags.ml", lines 551-556, characters 23-31:
Error (warning 8 [partial-match]): this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
Lambda
File "utils/clflags.ml", lines 559-563, characters 26-38:
Error (warning 8 [partial-match]): this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
Lambda
File "utils/clflags.ml", lines 575-579, characters 4-65:
Error (warning 8 [partial-match]): this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
Lambda
$ dune build --workspace=duneconf/boot.ws _build/default/.ocamlcommon.objs/native/warnings.cmx
File "utils/warnings.ml", line 545, characters 12-13:
Error: This expression has type int
       This is not a function; it cannot be applied.
```

ccasin:
@ncik-roberts yes, thanks, fixed.  I'm going to add a note to Stephen's instructions about remembering to check this for every file you think is "done" before PRing.

ncik-roberts (approved):
Thanks, now it LGTM for a more stringent definition of "looks".

PR #174, opened 2023-10-18
Fix backend conflicts

xclerc:
Covers most conflicts under `asmcomp/`;
@mshinwell is working on the _cmm_ ones.

(Very much untested.)


asmcomp/asmlink.ml
```diff
@@ -49,40 +49,15 @@ let cmx_required = ref ([] : CU.t list)
 
 let check_consistency file_name unit crc =
   begin try
```
gretay-js, line 51: There is a check upstream that is missing here, is it intentional? I think it's needed for L60 change below, see https://github.com/ocaml/ocaml/pull/11635.

```suggestion
  begin try
    let source = List.assoc unit.ui_name !implementations_defined in
    raise (Error(Multiple_definition(unit.ui_name, file_name, source)))
  with Not_found -> ()
  end;
  begin try
```

asmcomp/selectgen.ml
```diff
@@ -793,22 +773,12 @@ method emit_expr_aux (env:environment) exp :
           self#insert_debug env  (Iraise k) dbg rd [||];
           None
       end
-<<<<<<< HEAD
   | Cop(Ccmpf _, _, dbg) ->
```
gretay-js, line 776: just a note that the case of `Cop(Ccmpf _, _, dbg)`  doesn't exist in flambda-backend because we have already backported https://github.com/ocaml/ocaml/pull/9945.

gretay-js (approved):

gretay-js:
My review is for changes under `asmcomp` except `cmm*`.

asmcomp/asmlink.ml
lukemaurer: Looks like the check is still there, only moved to line 90: https://github.com/ocaml-flambda/ocaml-jst/blob/60d0234646097c947f5bed057a9f9d51a2eaa5c7/asmcomp/asmlink.ml#L90

(Not sure why the check is in a different place. Probably doesn't have to be.)

gretay-js: The upstream PR#11635 moves the check from L90 to L60 because otherwise the  now-unguarded crc check will give strange errors.

lukemaurer: Ah, good catch. I guess we should move it here, then? (Actually, how did the new version of the check end up further down to begin with? Did we backport that PR but move the check?)

mshinwell: I've moved the check up, @lukemaurer please confirm

PR #178, opened 2023-10-18
Fix stdlib mlis

ncik-roberts:
Resolve conflicts in stdlib mlis. These are almost always resolutions that prefer `tip-5`, but a handful of cases incorporate comments/documentation from `tip-jst`.

I make one update to `gc.ml` because, while we're still using the ocaml 4 runtime, we don't have to replace these functions with no-ops. When we finish merging the ocaml 5 runtime, we'll be flagged that these external functions don't exist and can figure out that they should be replaced with no-ops then.

Testing:

```
for x in stdlib/*.mli; do
  file=$(basename ${x%.mli});
  if echo $file | grep 'camlinternal\|stdlib\|std_exit'; then
    cmi=$file
  else
    cmi=stdlib__${file^}
  fi

  dune b --workspace=duneconf/runtime_stdlib.ws _build/runtime_stdlib/stdlib/.stdlib.objs/byte/$cmi.cmi
done
```

You'll see some syntax errors for JS-internal features: `local_`, `float#`, `global_`, but things otherwise build

### DIFF
--- a/stdlib/gc.mli
+++ b/stdlib/gc.mli
@@ -95,26 +95,16 @@ type stat =
 
     stack_size: int;
     (** Current size of the stack, in words.
-<<<<<<< HEAD
-        @since 3.12.0 *)
-||||||| merged common ancestors
-    (** Current size of the stack, in words. @since 3.12.0 *)
-=======
-        This metrics is currently not available in OCaml 5: the field value is
-        always [0].
+
+        This metrics will not be available in the OCaml 5 runtime: the field
+        value will always be [0].
+
         @since 3.12 *)
->>>>>>> ocaml/5.1
 
     forced_major_collections: int;
     (** Number of forced full major collections completed since the program
         was started.
-<<<<<<< HEAD
-        @since 4.12.0 *)
-||||||| merged common ancestors
-        was started. @since 4.12.0 *)
-=======
         @since 4.12 *)
->>>>>>> ocaml/5.1
 }
 (** The memory management counters are returned in a [stat] record. These
    counters give values for the whole program.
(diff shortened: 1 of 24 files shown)


stdlib/gc.mli
```diff
@@ -449,7 +433,6 @@ val delete_alarm : alarm -> unit
 (** [delete_alarm a] will stop the calls to the function associated
    to [a]. Calling [delete_alarm a] again has no effect. *)
 
-<<<<<<< HEAD
```
goldfirere, line 452: Why keep our code here? Presumably for compatibility with the ocaml 4 runtime. But then we should probably leave a `CR ocaml 5 runtime` (?) or something to come back and fix this later.

goldfirere:
Reviewing

stdlib/moreLabels.mli
```diff
@@ -1260,19 +1122,9 @@ module Set : sig
...
-         to the ordering [Ord.compare], where [Ord] is the argument
-         given to {!Stdlib.Set.Make}. *)
-||||||| merged common ancestors
-         The returned list is sorted in increasing order with respect
-         to the ordering [Ord.compare], where [Ord] is the argument
-         given to {!Make}. *)
-=======
           The returned list is sorted in increasing order with respect
           to the ordering [Ord.compare], where [Ord] is the argument
           given to {!Set.Make}. *)
```
goldfirere, line 1127:
```suggestion
          given to {!Stdlib.Set.Make}. *)
```

goldfirere (approved):

stdlib/gc.mli
goldfirere: Sounds good to me.

PR #171, opened 2023-10-18
Resolve conflicts in Lambda, Printlambda, Simplif and Tmc

mshinwell:
Can't check the build yet due to conflicts in cmi files in `typing/`

PR #179, opened 2023-10-18
Translation: Resolve conflicts in `matching.ml` and `translcore.ml`

ncik-roberts:
Cherry-picks a commit from #171; once that's merged I'll drop the `lambda.ml` and `lambda.mli` changes from here.

Nature of the resolutions:
  * (~60% of the conflicts) Upstream added an argument `Immediate | Pointer` to `Pfield` in addition to `mutable_flag`. We added `field_read_semantics` internally (this is semantically equivalent to `mutable_flag`, though the constructor names are different). The conflicts are easy to resolve: I just figure out `immediate_or_pointer` based on the upstream piece of the diff.
  * (~3 conflicts) Lazy field accesses are factored out to a function upstream; I add an extra argument to this function because we populated `ap_inlined` differently at one callsite (though I doubt it actually matters, as I explain in a comment).
  * The rest mostly have to do with modes/layouts. 

### Testing

```
$ dune b --workspace=duneconf/boot.ws _build/default/.ocamlcommon.objs/native/translcore.cmx
$ dune b --workspace=duneconf/boot.ws _build/default/.ocamlcommon.objs/native/matching.cmx
File "lambda/matching.ml", line 1990, characters 56-71:
Error: Unbound value Obj.forcing_tag
Hint: Did you mean string_tag?
```

To see that the latter is ok, I recommend adding this module to the top of the file:

```
module Obj = struct
  include Obj
  let forcing_tag = 0
end
```

Then it should build.

### DIFF
--- a/lambda/matching.ml
+++ b/lambda/matching.ml
@@ -1806,16 +1806,9 @@ let get_expr_args_constr ~scopes head (arg, _mut, sort, layout) rem =
       if pos > last_pos then
         argl
       else
-<<<<<<< HEAD
-        (Lprim (Pfield (pos, Reads_agree), [ arg ], loc), binding_kind,
+        (Lprim (Pfield (pos, Pointer, Reads_agree), [ arg ], loc), binding_kind,
          Jkind.Sort.for_constructor_arg, layout_field)
         :: make_args (pos + 1)
-||||||| merged common ancestors
-        (Lprim (Pfield pos, [ arg ], loc), binding_kind) :: make_args (pos + 1)
-=======
-        (Lprim (Pfield (pos, Pointer, Immutable), [ arg ], loc),
-               binding_kind) :: make_args (pos + 1)
->>>>>>> ocaml/5.1
     in
     make_args first_pos
   in
@@ -1841,21 +1834,15 @@ let divide_constructor ~scopes ctx pm =
 let get_expr_args_variant_constant = drop_expr_arg
 
 let nonconstant_variant_field index =
-  Lambda.Pfield(index, Reads_agree)
+  Lambda.Pfield(index, Pointer, Reads_agree)
 
 let get_expr_args_variant_nonconst ~scopes head (arg, _mut, _sort, _layout)
       rem =
   let loc = head_loc ~scopes head in
-<<<<<<< HEAD
    let field_prim = nonconstant_variant_field 1 in
   (Lprim (field_prim, [ arg ], loc), Alias, Jkind.Sort.for_constructor_arg,
    layout_field)
   :: rem
-||||||| merged common ancestors
-  (Lprim (Pfield 1, [ arg ], loc), Alias) :: rem
-=======
-  (Lprim (Pfield (1, Pointer, Immutable), [ arg ], loc), Alias) :: rem
->>>>>>> ocaml/5.1
 
 let divide_variant ~scopes row ctx { cases = cl; args; default = def } =
   let rec divide = function
@@ -1924,14 +1911,8 @@ let get_mod_field modname field =
   lazy
     (let mod_ident = Ident.create_persistent modname in
      let env =
-<<<<<<< HEAD
        Env.add_persistent_structure mod_ident
-         (Lazy.force Env.initial_safe_string)
-||||||| merged common ancestors
-       Env.add_persistent_structure mod_ident Env.initial_safe_string
-=======
-       Env.add_persistent_structure mod_ident Env.initial
->>>>>>> ocaml/5.1
+         (Lazy.force Env.initial)
      in
      match Env.open_pers_signature modname env with
      | Error `Not_found ->
@@ -1957,14 +1938,7 @@ let code_force_lazy = get_mod_field "CamlinternalLazy" "force_gen"
    Forward(val_out_of_heap).
 *)
 
-<<<<<<< HEAD
-let lazy_forward_field = Lambda.Pfield (0, Reads_vary)
-
-let inline_lazy_force_cond arg pos loc =
-||||||| merged common ancestors
-let inline_lazy_force_cond arg loc =
-=======
-let call_force_lazy_block varg loc =
+let call_force_lazy_block ?(inlined = Default_inlined) varg loc ~pos =
   (* The argument is wrapped with [Popaque] to prevent the rest of the compiler
      from making any assumptions on its contents (see comments on
      [CamlinternalLazy.force_gen], and discussions on PRs #9998 and #10909).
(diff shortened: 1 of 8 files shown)


goldfirere:
Reviewing

lambda/translcore.ml
```diff
@@ -1543,27 +1386,8 @@ and transl_function ~in_new_scope ~scopes e alloc_mode param arg_mode arg_sort r
...
   let loc = of_location ~scopes e.exp_loc in
-<<<<<<< HEAD
   let body = if region then maybe_region_layout return body else body in
   let lam = lfunction ~kind ~params ~return ~body ~attr ~loc ~mode ~region in
-||||||| merged common ancestors
-  let lam = Lfunction{kind; params; return; body; attr; loc} in
-  Translattribute.add_function_attributes lam e.exp_loc e.exp_attributes
-=======
-  let lam = lfunction ~kind ~params ~return ~body ~attr ~loc in
-  let attrs =
```
goldfirere, line 1554: This bit looks important. Is there a reason to drop?

goldfirere (approved):

lambda/matching.ml
```diff
@@ -2000,67 +1978,19 @@ let inline_lazy_force_cond arg loc =
...
-=======
                 (
                   (* ... if tag == Obj.lazy_tag || tag == Obj.forcing_tag then
                          Lazy.force varg
                        else ... *)
                   Lprim (Psequor,
                        [test_tag Obj.lazy_tag; test_tag Obj.forcing_tag], loc),
-                  call_force_lazy_block varg loc,
->>>>>>> ocaml/5.1
+                  (* nroberts: We probably don't need [Never_inlined] anymore
```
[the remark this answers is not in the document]
mshinwell: I'll look at this and the one below

lambda/translcore.ml
ncik-roberts: It's done at the top of the function internally.

lambda/matching.ml
```diff
@@ -2132,23 +2031,21 @@ let inline_lazy_force arg pos loc =
       { ap_tailcall = Default_tailcall;
         ap_loc = loc;
         ap_func = Lazy.force code_force_lazy;
-<<<<<<< HEAD
-        ap_args = [ arg ];
+        ap_args = [ Lconst (Const_base (Const_int 0)); arg ];
         ap_result_layout = Lambda.layout_lazy_contents;
         ap_region_close = pos;
         ap_mode = alloc_heap;
+        (* nroberts: To make sure this wasn't inlined:
```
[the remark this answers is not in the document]
ncik-roberts: My intention was to leave this as a comment for now and not have investigation block this PR.

lambda/translcore.ml
ncik-roberts: (We backported this and changed it slightly to make it work with zero-alloc attributes.)

lambda/translcore.ml
```diff
@@ -997,16 +891,8 @@ and transl_exp0 ~in_new_scope ~scopes sort e =
...
-<<<<<<< HEAD
               Llet(Alias, Lambda.layout_module_field, id,
                    Lprim(mod_field pos, [Lvar oid],
-||||||| merged common ancestors
-              Llet(Alias, Pgenval, id,
-                   Lprim(Pfield pos, [Lvar oid],
-=======
-              Llet(Alias, Pgenval, id,
-                   Lprim(Pfield (pos, Pointer, Mutable), [Lvar oid],
->>>>>>> ocaml/5.1
```
ncik-roberts, line 1009: I'm leaving this comment to reflect the fact we're discussing this internally. I'm surprised that upstream has `Mutable` but that `mod_field` is implemented with `Reads_agree` -- shouldn't it be `Reads_vary`?

ncik-roberts (commented):
leaving an unresolved conversation for internal thread

lambda/translcore.ml
ncik-roberts: cc @mshinwell for context on internal conversation

lambda/matching.ml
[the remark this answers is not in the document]
goldfirere: Very reasonable. But are we sure we'll know to look for this comment? Maybe a `CR ocaml 5 merge`? Maybe that's splitting hairs too much?

goldfirere:
#179 was merged into this. Only the first commit needs to be reviewed here.

lambda/translcore.ml
lthls: I haven't seen the internal conversation, but I assume that it revolves around recursive modules. In which case, I'll point to a previous analysis I did when reviewing a PR on recursive modules: https://github.com/ocaml/ocaml/pull/10205#pullrequestreview-674345642.
My opinion here is that using immutable reads (`Reads_agree`) should be preferred even if it could lead to observable differences:
- For functions, only physical equality can tell the difference (the stub function will behave as the new function if the module has been initialized)
- For lazy values, the old value could leak. This only matters if the lazy value was forced before initialization but the corresponding exception was caught and ignored; otherwise, the old value and the new value behave the same way after initialization.
- For classes, I think the old value could leak, but I haven't done enough tests to know whether it is actually a problem (the language doesn't allow to manipulate class values directly, so it's a bit hard to trigger the complex cases).
- For submodules, no observable difference for the modules themselves, as the modules fields are not patched. The fields of the submodules can be affected though.

lthls: >     * For classes, I think the old value could leak, but I haven't done enough tests to know whether it is actually a problem (the language doesn't allow to manipulate class values directly, so it's a bit hard to trigger the complex cases).

Update: I just checked, and classes are actually patched in place, so the old value and the new value are the same.

ncik-roberts: OK, so it sounds like it's fine to do `Reads_agree` here (which is what the code currently does). I'll resolve this, thanks. (Though I suspect others may occasionally run into this confusion as long as the Jane Street branch disagrees with `ocaml/ocaml` &mdash; should something be done to make them agree?)

lambda/matching.ml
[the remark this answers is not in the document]
ncik-roberts: As with https://github.com/ocaml-flambda/ocaml-jst/pull/179#discussion_r1365343946, my intention was to leave this as a comment for now and not have investigation block this PR. I'll resolve this.

ncik-roberts:
Thanks all, merging (not squashing as 1 of the 2 commits in this PR corresponds to a stacked PR).

ncik-roberts:
(Actually, looks like rebase and merge is disabled and it's not so bad to squash, so I'll do that.)

PR #180, opened 2023-10-18
Ocaml 5 runtime and configure tweaks

stedolan:
This PR tweaks the dune files so that the following two targets now build (the latter, with a suitable ocamlopt on PATH):

```
dune b --workspace=duneconf/boot.ws _build/default/.ocamlcommon.objs/native/domainstate.cmx
dune b --workspace=duneconf/runtime_stdlib.ws _build/runtime_stdlib/runtime4/libasmrun.a
```

There is also a tweak to configure.ac to re-enable stack overflow detection (which is done in a different way on the ocaml 5 runtime)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2069,6 +2069,17 @@ AS_CASE([$enable_debug_runtime],
   [no], [debug_runtime=false],
   [debug_runtime=true])
 
+## Determine if system stack overflows can be detected
+
+AC_MSG_CHECKING([whether stack overflows can be detected])
+
+AS_CASE([$arch,$system],
+  [i386,linux_elf|amd64,linux|amd64,macosx \
+    |amd64,openbsd|i386,bsd_elf|arm64,linux|arm64,macosx],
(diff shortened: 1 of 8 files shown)


configure.ac
```diff
@@ -2069,6 +2069,17 @@ AS_CASE([$enable_debug_runtime],
...
+
+AC_MSG_CHECKING([whether stack overflows can be detected])
+
+AS_CASE([$arch,$system],
+  [i386,linux_elf|amd64,linux|amd64,macosx \
+    |amd64,openbsd|i386,bsd_elf|arm64,linux|arm64,macosx],
+    [AC_DEFINE([HAS_STACK_OVERFLOW_DETECTION])
+    AC_MSG_RESULT([yes])],
+  [AC_MSG_RESULT([no])])
+
```
stedolan, line 2082: This shouldn't be merged until someone has run `autoconf` and committed the new `configure` script on a machine with recent autoconf installed.

mshinwell (approved):

PR #182, opened 2023-10-19
Translation: resolve conflicts in remaining files (`translmod`, `transclass`, `translobj`, `translprim`, `translattribute`, `switch`)

ncik-roberts:
Finish off what #179 and #171 started.

Notable resolutions:
* As explained in #179, there's a new field on `Pfield`. An unresolved thread on that PR is also relevant here: Should accessing a module field be `Reads_vary` like it is upstream? I've asked Mark to weigh in.
* `Lambda.lfunction` was made a private type with a smart constructor both internally and upstream, and we have some extra arguments (region, mode). These are trivial to resolve.
* Two small diffs that we pull in from: https://github.com/ocaml-flambda/flambda-backend/pull/871
  * explains diff in `translmod.ml` where we don't need `is_compilation_unit`
  * explains diff in `translprim.ml` for `add_used_primtive` (where a change made upstream is irrelevant for us)
* Hardcoded checks for misplaced attribute warning go away internally, in preference for the more general framework in Chris's patch
* Threading through of kinds

### Testing

```
for module in translclass translobj translmod translprim translattribute switch; do
  dune b --workspace=duneconf/boot.ws _build/default/.ocamlcommon.objs/native/$module.cmx
done
```

### DIFF
--- a/lambda/translattribute.ml
+++ b/lambda/translattribute.ml
@@ -50,10 +50,6 @@ let is_poll_attribute =
 let is_loop_attribute =
   [ ["loop"; "ocaml.loop"], true ]
 
-let is_poll_attribute = function
-  | {txt=("poll")} -> true
-  | _ -> false
-
 let find_attribute p attributes =
   let inline_attribute = Builtin_attributes.filter_attributes p attributes in
   let attr =
@@ -249,7 +245,6 @@ let parse_local_attribute attr =
         ]
         payload
 
-<<<<<<< HEAD
 let parse_property_attribute attr property =
   match attr with
   | None -> Default_check
@@ -293,18 +288,6 @@ let parse_loop_attribute attr =
         [
           "never", Never_loop;
           "always", Always_loop;
-||||||| merged common ancestors
-=======
-let parse_poll_attribute attr =
-  match attr with
-  | None -> Default_poll
-  | Some {Parsetree.attr_name = {txt; loc}; attr_payload = payload} ->
-      parse_id_payload txt loc
-        ~default:Default_poll
-        ~empty:Default_poll
-        [
-          "error", Error_poll;
->>>>>>> ocaml/5.1
         ]
         payload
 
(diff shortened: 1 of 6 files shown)


lambda/translattribute.ml
```diff
@@ -638,50 +518,6 @@ let add_poll_attribute expr loc attributes =
...
-   appears on any expression, it is an error, otherwise it would
-   have been removed by this function *)
-let get_and_remove_inlined_attribute e =
-  let attr, exp_attributes =
-    find_attribute is_inlined_attribute e.exp_attributes
-  in
-  let inlined = parse_inline_attribute attr in
-  inlined, { e with exp_attributes }
-=======
-  match expr, get_poll_attribute attributes with
```
goldfirere, line 654: I'm pretty sure this upstream code behaves the same as our version. Might as well use the upstream version to avoid future conflicts.

lambda/translattribute.ml
goldfirere, line 1: It looks like we don't support the `poll` attribute. Is that the intent?

goldfirere (approved):

lambda/translattribute.ml
ncik-roberts: It doesn't show up in the diff but it's a little higher up:

```
let is_poll_attribute =
  [ ["poll"; "ocaml.poll"], true ]
```

that's what we want.

lambda/translattribute.ml
ncik-roberts: I think the internally-adopted style is actually necessary for the misplaced attribute warning to work.

See Chris's commit here: https://github.com/ocaml-flambda/ocaml-jst/commit/a9268d29b47cd5c3c051e761f454330573f7ee0e

The conflict will be resolved when he finishes upstreaming that.

goldfirere:
Hmmm.... maybe that was a bad idea because this was targeting a branch.

ncik-roberts:
I think that's OK. There's only two unresolved small questions on the target that Mark will look at tomorrow, and I don't think this is blocking anything (it's just mls).

PR #187, opened 2023-10-19
OCaml 5: resolve remaining mli conflicts

ncik-roberts:
Finish resolving all mli conflicts. I'm not sure if anyone is blocked on these files in particular, but it's nice to just be done resolving mlis so almost everything can be built.

Testing:

I couldn't test `toplevel/topcommon.mli`, `otherlibs/systhreads/thread.mli`, or `ocamltest/ocamltest_config.ml`, as these are built with another dune profile that relies on the compiler being able to build the stdlib. I think the conflict resolutions are simple enough that we should just merge them as-is.

```
# asmcomp/mach.mli
$ dune b --workspace=duneconf/boot.ws _build/default/.ocamloptcomp.objs/byte/mach.cmi

# bytecomp/bytelink.mli
$ dune b --workspace=duneconf/boot.ws _build/default/.ocamlbytecomp.objs/byte/bytelink.cmi

# middle_end/flambda/flambda_utils.mli
$ dune b --workspace=duneconf/boot.ws _build/default/.ocamloptcomp.objs/byte/flambda_utils.cmi
```

### DIFF
--- a/middle_end/flambda/flambda_utils.mli
+++ b/middle_end/flambda/flambda_utils.mli
@@ -69,13 +69,9 @@ val make_closure_declaration
   -> region:bool
   -> body:Flambda.t
   -> params:Parameter.t list
-<<<<<<< HEAD
   -> return_layout:Lambda.layout
   -> free_variables:Lambda.layout Variable.Map.t
-||||||| merged common ancestors
   -> stub:bool
-=======
->>>>>>> ocaml/5.1
   -> Flambda.t
 
 val toplevel_substitution
--- a/middle_end/flambda/flambda_utils.mli
+++ b/middle_end/flambda/flambda_utils.mli
@@ -71,7 +71,6 @@ val make_closure_declaration
   -> params:Parameter.t list
   -> return_layout:Lambda.layout
   -> free_variables:Lambda.layout Variable.Map.t
-  -> stub:bool
   -> Flambda.t
 
 val toplevel_substitution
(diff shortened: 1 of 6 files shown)


middle_end/flambda/flambda_utils.mli
```diff
@@ -69,13 +69,9 @@ val make_closure_declaration
   -> region:bool
   -> body:Flambda.t
   -> params:Parameter.t list
-<<<<<<< HEAD
   -> return_layout:Lambda.layout
   -> free_variables:Lambda.layout Variable.Map.t
-||||||| merged common ancestors
   -> stub:bool
```
goldfirere, line 74: Looks like both we and they removed this parameter. Drop it?

ocamltest/ocamltest_config.mli
```diff
@@ -138,7 +128,6 @@ val function_sections : bool
 val instrumented_runtime : bool
 (** Whether the instrumented runtime is available *)
 
-<<<<<<< HEAD
 val naked_pointers : bool
```
goldfirere, line 131: This will get removed with the ocaml 5 runtime, I think. In any case, upstream removed it.

goldfirere (approved):

middle_end/flambda/flambda_utils.mli
ncik-roberts: Good catch, thanks. That was a mistake.

ocamltest/ocamltest_config.mli
ncik-roberts: I wasn't sure whether we'd want to keep it around until we've gotten around to deleting the flag -- I'll leave this as a `CR ocaml 5 runtime`.

PR #189, opened 2023-10-19
Build OCaml 5 files in CI

ncik-roberts:
Add two things:
  * A script that tries to build individual ml/mli files. The intention is that you can edit it in your PR to (1) provide evidence to your reviewer that your changes work, and (2) preserve that evidence for future runners of the script.
  * A CI job that runs the script.

I think the first thing is very useful and the second thing is marginally useful.

Rigging up CI was tricky because it requires menhir in addition to the dune/OCaml setup from existing CI jobs.

The builds of these things (OCaml, dune, menhir) are cached, so it should take less than a minute per CI run. (See [this run](https://github.com/ocaml-flambda/ocaml-jst/actions/runs/6580127292/job/17877459123) for evidence of the caching.)

But, if we find that the CI job takes too long to run and is blocking merging, I think we should get rid of the job. The script is useful enough on its own. I'll add documentation for it in a common place (+ nudge people to use it).

### DIFF
--- a/.github/workflows/jane_ocaml5.yml
+++ b/.github/workflows/jane_ocaml5.yml
@@ -13,11 +13,54 @@ jobs:
         working-directory: ocaml-jst
         run: |
           ./configure --prefix=$GITHUB_WORKSPACE/ocaml-5/_install
+      - name: Cache OCaml 4.14 and dune
+        uses: actions/cache@v2
+        id: cache
+        with:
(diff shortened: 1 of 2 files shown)


ccasin (approved):
Clearly nice to have, since it means reviewers don't have to manually check that things build.

I took a quick look at the script and yml file, which look fine, but it doesn't seem worth reviewing very carefully since bugs here are low consequence and we'll find out experimentally if there are problems.

PR #190, opened 2023-10-20
OCaml5 merge conflicts in `tast_{mapper/iterator}.ml`

ccasin:
One thing to consider: Upstream added the ability to iterate over locations and attributes.  So, anywhere we've added something to the typed tree that contains locations or attributes, we must add a call to the appropriate iterators.  I think the only such places are in comprehensions.

Reviewer: @lpw25

### DIFF
--- a/typing/tast_mapper.ml
+++ b/typing/tast_mapper.ml
@@ -143,9 +143,9 @@ let include_kind sub = function
 let str_include_infos sub x =
   let incl_loc = sub.location sub x.incl_loc in
   let incl_attributes = sub.attributes sub x.incl_attributes in
-  { x with incl_loc; incl_attributes;
-           incl_mod = sub.module_expr sub x.incl_mod;
-           incl_kind = include_kind sub x.incl_kind }
+  let incl_mod = sub.module_expr sub x.incl_mod in
+  let incl_kind = include_kind sub x.incl_kind in
+  { x with incl_loc; incl_attributes; incl_mod; incl_kind }
 
 let class_type_declaration sub x =
   class_infos sub (sub.class_type sub) x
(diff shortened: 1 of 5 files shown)


typing/tast_mapper.ml
```diff
@@ -146,16 +141,11 @@ let include_kind sub = function
...
-let include_infos f x = {x with incl_mod = f x.incl_mod}
-=======
-let include_infos sub f x =
   let incl_loc = sub.location sub x.incl_loc in
   let incl_attributes = sub.attributes sub x.incl_attributes in
-  {x with incl_loc; incl_attributes; incl_mod = f x.incl_mod}
->>>>>>> ocaml/5.1
+  { x with incl_loc; incl_attributes;
+           incl_mod = sub.module_expr sub x.incl_mod;
+           incl_kind = include_kind sub x.incl_kind }
```
lpw25, line 148: Would you mind just making the styles match here?

typing/tast_mapper.ml
```diff
@@ -631,7 +557,10 @@ let signature sub x =
   {x with sig_items; sig_final_env}
 
 let sig_include_infos sub x =
-  { x with incl_mod = sub.module_type sub x.incl_mod;
+  let incl_loc = sub.location sub x.incl_loc in
+  let incl_attributes = sub.attributes sub x.incl_attributes in
+  { x with incl_loc; incl_attributes;
+           incl_mod = sub.module_type sub x.incl_mod;
            incl_kind = include_kind sub x.incl_kind }
```
lpw25, line 564: Also here

lpw25 (approved):
A couple of points where the merge creates mismatched styles. Other than that, good to merge.

PR #192, opened 2023-10-20
Resolve conflicts in ocamltest/

mshinwell:
As subject


ocamltest/ocaml_compilers.ml
```diff
@@ -40,30 +40,14 @@ class compiler
...
-    Builtin_variables.program
-||||||| merged common ancestors
-    if Ocaml_backends.is_native host
-    then Builtin_variables.program2
-    else Builtin_variables.program
-=======
     if Ocaml_backends.is_native host && not Sys.win32
     then Builtin_variables.program2
     else Builtin_variables.program
->>>>>>> ocaml/5.1
```
ncik-roberts, line 53: Why not resolve with `tip-jst`? It looks like we changed it to just `Builtin_variables.program` in https://github.com/ocaml-flambda/flambda-backend/pull/504.

ocamltest/ocaml_compilers.ml
```diff
@@ -40,30 +40,14 @@ class compiler
...
-    Some Builtin_variables.output
-||||||| merged common ancestors
-    if Ocaml_backends.is_native host
-    then None
-    else Some Builtin_variables.output
-=======
     if Ocaml_backends.is_native host && not Sys.win32
     then None
     else Some Builtin_variables.output
->>>>>>> ocaml/5.1
```
ncik-roberts, line 66: Same question as above.

ocamltest/ocaml_tests.ml
```diff
@@ -20,51 +20,6 @@ open Builtin_actions
...
 let bytecode =
-<<<<<<< HEAD
-  let test_actions =
-    if not Ocamltest_config.native_compiler then
-      [
-        setup_ocamlc_byte_build_env;
-        ocamlc_byte;
-        check_ocamlc_byte_output;
-        run;
-        check_program_output;
```
ncik-roberts, line 31: In `tip-jst`, we would not run these actions when `Ocamltest_config.native_compiler` is true:

```
        setup_ocamlc_byte_build_env;
        ocamlc_byte;
        check_ocamlc_byte_output;
```
(And this seems like an intentional divergence from upstream; see https://github.com/ocaml-flambda/flambda-backend/pull/504.)

In the resolved code, we do run those actions when `Ocamltest_config.native_compiler` is true. Should we try to fix that here?

ocamltest/ocaml_tests.ml
```diff
@@ -96,34 +51,8 @@ let bytecode =
...
-    setup_ocamlopt_byte_build_env;
-    ocamlopt_byte;
-    check_ocamlopt_byte_output;
-    run;
-    check_program_output;
-    setup_ocamlopt_opt_build_env;
-    ocamlopt_opt;
-    check_ocamlopt_opt_output;
-  ] in
-=======
```
ncik-roberts, line 126: Same comment above -- should we try to preserve https://github.com/ocaml-flambda/flambda-backend/pull/504 in our resolution?

ncik-roberts (commented):
A few question around whether we should try to preserve behavior from https://github.com/ocaml-flambda/flambda-backend/pull/504.

Also if you figured out how to build these, mind adding those to the CI script?

ocamltest/ocaml_compilers.ml
mshinwell: Ah yes, indeed.  Fixed

ocamltest/ocaml_compilers.ml
mshinwell: Fixed

ocamltest/ocaml_tests.ml
mshinwell: It seems like an equivalent of PR504 (but not involving the changes to `ocaml_compilers.ml`) was already merged in 5.1 but enabled only on Windows.  I've now enabled it unconditionally to match PR504, and have put back in the change in `ocaml_compilers.ml` from that PR.

ocamltest/ocaml_tests.ml
mshinwell: As above

mshinwell:
Build script also updated to build ocamltest.

ocamltest/ocaml_tests.ml
ncik-roberts: Thanks

PR #193, opened 2023-10-20
OCaml5 conflicts: typedecl and typeclass

ccasin:
Not much notable here.  Some annoying conflicts in typedecl caused by a couple upstream PRs which we backported but were then built on a bit more upstream.

Review: @lpw25

### DIFF
--- a/typing/typedecl.ml
+++ b/typing/typedecl.ml
@@ -2491,8 +2491,8 @@ let transl_with_constraint id ?fixed_row_path ~sig_env ~sig_decl ~outer_env
 
 let abstract_type_decl ~injective jkind params =
   let arity = List.length params in
-  let params = List.map Ctype.newvar params in
   Ctype.with_local_level ~post:generalize_decl begin fun () ->
+    let params = List.map Ctype.newvar params in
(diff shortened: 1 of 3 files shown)


typing/typedecl.ml
```diff
@@ -2881,27 +2489,11 @@ let transl_with_constraint id ?fixed_row_path ~sig_env ~sig_decl ~outer_env
 
 (* Approximate a type declaration: just make all types abstract *)
 
-<<<<<<< HEAD
 let abstract_type_decl ~injective jkind params =
   let arity = List.length params in
-  Ctype.begin_def();
   let params = List.map Ctype.newvar params in
```
lpw25, line 2494: This `newvar` looks like it has moved from within the `begin_def` to outside of it.

lpw25 (commented):
Looks like one small unintentional change that could be a bug. The rest looks good.

typing/typedecl.ml
ccasin: well spotted

lpw25 (approved):

PR #196, opened 2023-10-23
OCaml5 conflicts: `typing/typemod.ml`

ccasin:
The main causes of conflicts were:
- include functor
- Places where we backported shape changes on lines that also have laziness or strengthening changes.

Review: @lpw25


lpw25 (approved):
LGTM

PR #197, opened 2023-10-23
Resolving conflicts in bytecomp/

mshinwell:
@lukemaurer please could you review this, and check carefully - it's likely I've made mistakes here.  One thing you might want to do is to diff the files on this branch directly from the `tip-5` tag, rather than looking directly at this PR's diff.  This shows our changes compared to upstream OCaml 5, so that diff should kind of look like the original symbols patch (plus some other minor modifications for other bytecode files which should be straightforward).  Unfortunately there were changes in OCaml 5 regarding the relocation processing, although it now seems to me that maybe they don't matter too much, since we don't need the renaming of relocations any more (iirc)?

### DIFF
--- a/bytecomp/bytelink.ml
+++ b/bytecomp/bytelink.ml
@@ -19,17 +19,14 @@ open Misc
 open Config
 open Cmo_format
 
-<<<<<<< HEAD
 module CU = Compilation_unit
-||||||| merged common ancestors
-=======
+
 module Dep = struct
   type t = string * string
   let compare = compare
 end
 
 module DepSet = Set.Make (Dep)
->>>>>>> ocaml/5.1
 
 type error =
   | File_not_found of filepath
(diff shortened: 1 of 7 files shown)


bytecomp/bytelink.ml
```diff
@@ -212,39 +192,24 @@ let implementations_defined = ref ([] : (CU.Name.t * string) list)
...
-            if name = cu.cu_name
-            then Consistbl.set crc_interfaces name crc file_name
-            else Consistbl.check crc_interfaces name crc file_name)
-=======
-        | Some crc -> Consistbl.check crc_interfaces name crc file_name)
->>>>>>> ocaml/5.1
+            (* CR mshinwell for lmaurer: please double-check, these two used
+               to be [Consistbl.set] and [Consistbl.check] respectively *)
+            then Consistbl.check crc_interfaces name full_name crc file_name
+            else Consistbl.check_noadd crc_interfaces name full_name crc file_name)
```
lukemaurer, line 212: I don't know any reason we can't just accept the upstream change here, namely just write
```ocaml
Consistbl.check crc_interfaces name full_name crc file_name
``` 
in place of the whole if/then/else. Notably, this is what #174 currently does: https://github.com/ocaml-flambda/ocaml-jst/blob/60d0234646097c947f5bed057a9f9d51a2eaa5c7/asmcomp/asmlink.ml#L60

lukemaurer (commented):
LGTM, save the one discrepancy with the other PR.

bytecomp/bytelink.ml
mshinwell: Fixed

lukemaurer (approved):

PR #199, opened 2023-10-23
OCaml 5: resolve conflicts in debugger

ncik-roberts:
Mostly trivial. The one decision was how to resolve the conflicts in the environment arg in `loadprinter.ml`, but https://github.com/ocaml-flambda/ocaml-jst/commit/ca305a2893c4c34930dae5f94271069f1bfe9976 makes it clear that we should use the empty argument as we're no longer parsing strings here.

Testing is in the build script. I had to fix a lingering `_intf.ml` merge conflict in dynlink to get things to build. Also, upstream merged `yacc/Makefile` into the main `Makefile`, so the dune build for yacc (which is used in some places in the debugger) needed a tweak.


ncik-roberts:
OK, actually I couldn't easily get the dune build to pass using the top-level Makefile, so I've restored the `yacc/Makefile` from `tip-jst` and added a comment to indicate we should think about deleting it.

debugger/loadprinter.ml
```diff
@@ -91,69 +91,14 @@ let eval_value_path env path =
...
-    (Ctype.instance desc.val_type);
-  Ctype.end_def();
-  Ctype.generalize ty_arg;
-  ty_arg
-=======
 let match_printer_type desc make_printer_type =
   Ctype.with_local_level ~post:Ctype.generalize begin fun () ->
-    let ty_arg = Ctype.newvar() in
+    let ty_arg = Ctype.newvar Jkind.(value ~why:Debug_printer_argument) in
     Ctype.unify Env.empty
```
poechsel, line 97: Shouldn't `Env.empty` be ` (Lazy.force Env.initial_safe_string)`?

ncik-roberts: I don't think so, I think https://github.com/ocaml-flambda/ocaml-jst/commit/ca305a2893c4c34930dae5f94271069f1bfe9976 suggests we should be using the empty environment. (That was an upstream PR that replaced uses of a full environment with the empty one, because we're no longer doing string lookups.)

ncik-roberts: Here's the commit upstream: https://github.com/ocaml/ocaml/commit/ca305a2893c4c34930dae5f94271069f1bfe9976

poechsel: Ack, sounds good then

poechsel (approved):

PR #191, opened 2023-10-20
Comment out contents of the Effect stdlib module and skip effect-related tests

mshinwell:
The effect primitives will also be made to raise fatal errors in `Translprim` in a separate PR.

PR #202, opened 2023-10-23
Resolve conflicts in testsuite/

mshinwell:
There turned out to be a lot of changes in generated Lambda code, backtraces, type error messages etc., which I've basically paid no attention to here in the majority of instances.  We can promote these once the compiler builds and we can run the testsuite properly.

There are however some proper changes mixed in here.


mshinwell:
@ncik-roberts is going to review

ncik-roberts:
Reviewing

testsuite/tests/backtrace/backtrace_effects_nested.ml
```diff
@@ -1,6 +1,4 @@
 (* TEST
-   * skip
-
```
ncik-roberts, line 3: Why delete this? That seems to contradict #191

ncik-roberts: I'll resolve by restoring this and similar, and let you object before merging

testsuite/tests/shapes/comp_units.ml
```diff
@@ -183,166 +182,6 @@ end
...
- "With_constraining_constraint"[module] ->
-   {<.16>
-    "M"[module] ->
-      {<.13>
-       "t"[type] ->
-         CU Stdlib . "Set"[module] . "Make"[module](
-         CU Stdlib . "Int"[module])<.13> . "t"[type];
-       };
-    };
->>>>>>> ocaml/5.1
```
ncik-roberts, line 345: Any reason we shouldn't just take `tip-5` here? (Do you know a reason why we wouldn't want to test this?)

testsuite/tests/shapes/functors.ml
```diff
@@ -238,229 +237,6 @@ module Big_to_small2 : B2S = functor (X : Big) -> struct include X end
...
-|}]
-
-module Big_to_small2 : B2S = functor (X : Big) -> struct include X end
-[%%expect{|
-{
- "Big_to_small2"[module] ->
-   Abs<.42>(X/390, {
-                    "t"[type] -> X/390<.41> . "t"[type];
-                    });
->>>>>>> ocaml/5.1
```
ncik-roberts, line 463: Same question as above.

ncik-roberts (approved):
I pushed a number of corrections &mdash; please take a look at them and let me know if you disagree with any of them.

We can consider this approved except for the unresolved question of whether we need to restore `skip` on the effect tests.

testsuite/tests/backtrace/backtrace_effects_nested.ml
mshinwell: I think that was just a mistake, these should be skipped.

mshinwell:
I think this is ok except for `931c20c148291c5de92d2357a8d9f46abdde104d ` which appears to have conflict markers on the + side of the diff?

mshinwell:
LGTM

PR #203, opened 2023-10-23
OCaml 5: resolve conflicts in stdlib mls

ncik-roberts:
Notes on the implementation:
  * The new function `filter_mapi` in `list.ml` is changed along the lines of our internal implementation to pass `f` as an argument rather than closing over it.
  * `stdlib/camlinternalOO.ml`: I inline the aliases to `Obj.*` functions, just like `tip-5`

Testing: I did not test, I think we'll have to wait until everything builds. David Allsopp's patches will be replayed on top of this to fix the incompatibilities with the OCaml 5 runtime, e.g. the unimplemented intrinsic `%atomic_cas`.

### DIFF
--- a/stdlib/camlinternalOO.ml
+++ b/stdlib/camlinternalOO.ml
@@ -131,13 +131,7 @@ let dummy_table =
 let table_count = ref 0
 
 (* dummy_met should be a pointer, so use an atom *)
-<<<<<<< HEAD
 let dummy_met : item = of_repr (Obj.new_block 0 0)
-||||||| merged common ancestors
-let dummy_met : item = obj (Obj.new_block 0 0)
-=======
-let dummy_met : item = Obj.obj (Obj.new_block 0 0)
->>>>>>> ocaml/5.1
 (* if debugging is needed, this could be a good idea: *)
 (* let dummy_met () = failwith "Undefined method" *)
 
--- a/stdlib/camlinternalOO.ml
+++ b/stdlib/camlinternalOO.ml
@@ -14,7 +14,6 @@
 (*                                                                        *)
 (**************************************************************************)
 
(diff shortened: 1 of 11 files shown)


ncik-roberts:
This just isn’t ready to review period.
It’s missing a commit. I’ll fix tomorrow.

stdlib/stdlib.ml
```diff
@@ -83,23 +83,9 @@ external ( != ) : ('a[@local_opt]) -> ('a[@local_opt]) -> bool = "%noteq"
 
 (* Boolean operations *)
 
-<<<<<<< HEAD
-external not : (bool[@local_opt]) -> bool = "%boolnot"
-external ( & ) : (bool[@local_opt]) -> (bool[@local_opt]) -> bool = "%sequand"
 external ( && ) : (bool[@local_opt]) -> (bool[@local_opt]) -> bool = "%sequand"
 external ( or ) : (bool[@local_opt]) -> (bool[@local_opt]) -> bool = "%sequor"
```
Ekdohibs, line 87: I think you removed `not` here where you wanted to remove `( or )` instead.

ncik-roberts: Indeed, thanks for catching that.

stdlib/camlinternalOO.ml
```diff
@@ -143,7 +132,7 @@ let table_count = ref 0
 
 (* dummy_met should be a pointer, so use an atom *)
 <<<<<<< HEAD
```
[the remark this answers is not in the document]
ncik-roberts: Now it is, preferring `tip-jst` as we wrap `Obj.repr` in `Sys.opaque_identity` to avoid inlining.

Ekdohibs (approved):

PR #205, opened 2023-10-24
Fix merge conflicts in api_docgen

poechsel:


ncik-roberts (approved):
All the resolutions look correct, though it's hard to be confident without testing. We'll do that testing later once we can build the compiler.

PR #209, opened 2023-10-24
Resolve conflicts in tools/

poechsel:
I wasn't able to add `dumpobj` to the ci script as it's failing to build due to some opcodes not being defined (opPERFORM/RESUME)


tools/dumpobj.ml
```diff
@@ -73,80 +73,6 @@ let record_events orig evl =
...
-          List.iter (fun a -> printf ", "; print_struct_const a) al;
-          printf ")"
-      end
-  | Const_float_array a ->
-      printf "[|";
-      List.iter (fun f -> print_float f; printf "; ") a;
-      printf "|]"
-
-=======
->>>>>>> ocaml/5.1
```
ncik-roberts, line 149: I was confused about this but it looks fine when I cross-reference https://github.com/ocaml-flambda/ocaml-jst/commit/2cff470317783a1483c19abcba4a2de034b5845f#diff-c81880f7e675038d344af33a9e0b1d364fe39f0c31715e9fab07658ba3afc765. The corresponding reading code in `bytecomp/emitcode.ml` is already updated with the corresponding other half of this diff.

ncik-roberts (approved):
Looks good to me. I added some more things to the build script (except for `dumpobj`, for the reason you mentioned about the new opcodes).

PR #213, opened 2023-10-24
OCaml5 merge conflicts: `typing/typecore.ml`

ccasin:
The beginning parts of this PR are not so bad, but towards the end I think it will be hard to review by reading the diff.  When looking over the completed version myself, I found the easiest thing was to first create a file containing the (pat)diffs between (this file and the old ocaml-jst version) and another with the diff between (this file and the upstream version).  Then, when reading a resolved diff, I would mainly look at those other files to understand how the diff was resolved and check nothing was dropped.  (Though I confess past a certain point I ran out of steam for carefully reading what I'd written - will continue looking tomorrow)

Some major sources of conflicts:

- 5.1 has some of Nick's cleanup work in typecore (https://github.com/ocaml/ocaml/pull/12117) but not all of it (https://github.com/ocaml/ocaml/pull/12229).  This is mainly in the checking of patterns, and results in conflicts when PRs in both categories change the same line, and what we have is already ahead of upstream 5.1.

- `type_pat` has been split into two functions upstream: `type_pat` and `check_counter_example_pat`.  These are meant to be kept in sync, but we didn't have the latter until now.  I've tried to make sure `check_counter_example_pat` has anything relevant we've added to `type_pat`, and we discussed the modes question elsewhere.

- Now that `type_pat` doesn't take a `~mode` argument to distingish it from `check_counter_example_pat`, it's very tempting to rename its `~alloc_mode` argument to `~mode`.  But it seemed like that would make for confusing diff here, so it should be done separately.

- We've completely rewritten type_argument internally, and there are upstream changes in it.  I proceeded by taking ours and then manually applying changes that had happened upstream.  It looks like there are three main upstream PRs that have touched this code:
  - https://github.com/ocaml/ocaml/pull/11536.  Mostly straightforward.  That upstream PR moves a level bump from the call site of type_application to a smaller part of its body, but it looks like we've already made a similar change internally.
  - https://github.com/ocaml/ocaml/pull/10364.  This factors out a function "type_unknown_args" for reasons that are unclear to me, but it appears a similar change had already been made in our implementation.
  - https://github.com/ocaml/ocaml/pull/11679..  This adds some location information in type_application that is used to improve an error message.  I believe I have replicated its behavior in our version - these locations were mostly already being tracked inside of `untyped_apply_arg`, I just needed to add an (optional) location in the `Eliminated_optional_arg` case.  If I got it wrong, the worst case is a marginally worse error, which we can easily detect experimentally later.

- type_let has been completely refactored upstream, and we've made a bunch of changes to it for layouts and modes.  Nothing too surprising here, just big diffs with confusing markers.


ncik-roberts:
I said I might merge this prior to Leo's review, but there a number of other open PRs we'll need before this is useful, so I'll hold off.

typing/typecore.ml
```diff
@@ -4382,8 +3390,28 @@ let collect_unknown_apply_args env funct ty_fun mode_fun rev_args sargs ret_tvar
...
+                  |> List.find_map
+                       (function
+                         | (_, Arg ( Known_arg { sarg; _ }
+                                   | Unknown_arg { sarg; _ })) ->
+                           Some sarg.pexp_loc
+                         | (_,
+                            Arg (Eliminated_optional_arg { next_arg_loc })) ->
+                           next_arg_loc
+                         | (_, Omitted _) -> None)
+                  |> Option.value ~default:funct.exp_loc
```
lpw25, line 3407: I think the upstream code here is just wrong. The comment is incorrect: the arguments are not in reverse order of appearance. I also think that `next_arg_loc` is not a sensible value here -- again this is also true of upstreams version. We should get rid of `next_arg_loc` now, and fix the rest later. Maybe leave a CR by the comment pointing out that it is wrong.

ncik-roberts (approved):
I didn't actually read this. I'm only approving so I can merge this now, even though Leo isn't done reviewing &mdash; he'll flag any further issues to Chris.

typing/typecore.ml
```diff
@@ -7923,83 +6226,34 @@ and type_function
...
-        | Label_mismatch { got; expected; expected_type} ->
-            Abstract_wrong_label { got; expected; expected_type; explanation }
-        | Not_a_function -> begin
-            match in_function with
-            | Some _ -> Too_many_arguments(ty_fun, explanation)
-            | None   -> Not_a_function(ty_fun, explanation)
-          end
-=======
-  let ty_arg, ty_res =
     with_local_level_iter_if separate ~post:generalize_structure begin fun () ->
```
lpw25, line 6237: This would probably be slightly easier to read if it used `with_local_level_if` and had an ad-hoc `post` function.

typing/typecore.ml
```diff
@@ -9408,10 +7127,9 @@ and type_construct env (expected_mode : expected_mode) loc lid sarg
               (instance ty_expected));
         end
       in
-      ((ty_args, ty_res, texp), ty_res::ty_args)
```
lpw25, line 9411: Maybe better to just use `with_local_level_if` and write our own post function. That `List.map` offends my sensibilities.

typing/typecore.ml
```diff
@@ -5579,30 +4574,50 @@ let vb_exp_constraint {pvb_expr=expr; pvb_pat=pat; pvb_constraint=ct; _ } =
...
+        Pat.constraint_ ~loc pat t
+    | None, (Ppat_any | Ppat_constraint _), _ -> pat
+    | None, _, Pexp_coerce (_, _, sty)
+    | None, _, Pexp_constraint (_, sty) when !Clflags.principal ->
+        (* propagate type annotation to pattern,
+           to allow it to be generalized in -principal mode *)
+        Pat.constraint_ ~loc:{pat.ppat_loc with Location.loc_ghost=true} pat sty
+    | _ -> pat
+  in
+  let pat_mode, exp_mode =
```
lpw25, line 4600: I'm not sure it makes much sense to pull this code out into this function. It seems to mix two separate concerns:
1. Converting constraints within value bindings into equivalent constraints on patterns.
2. Working out what modes the pattern and expression should be typed at

typing/typecore.ml
```diff
@@ -10929,85 +7816,36 @@ and type_andops env sarg sands expected_sort expected_ty =
         expected_sort,
         []
     | { pbop_op = sop; pbop_exp = sexp; pbop_loc = loc; _ } :: rest ->
-<<<<<<< HEAD
-        if !Clflags.principal then begin_def ();
-        let op_path, op_desc = type_binding_op_ident env sop in
-        let op_type = op_desc.val_type in
```
lpw25, line 10935: We've lost the removal of the call to `instance` on this line. The call is harmless but also not needed and potentially confusing since this is a difference between our `type_ident` and the one upstream. We should also probably remove it from the other call site of `type_binding_op_ident`.

lpw25 (commented):
I don't see any bugs but there are a few things that I think should be improved.

typing/typecore.ml
ccasin: I agree `next_arg_loc` is kind of nonsense and will get rid of it.

To check: I believe `rev_args` are in reverse order of appearance _in the type of the function_.  I think the point you are making is that may be different than the reverse order of the argument terms in the program due to labeled/optional argument.  Correct?

typing/typecore.ml
ccasin: Yeah, I waffled on this originally.  The advantage of this approach is that it saves iterating and allocating an additional list, but now the name is bad and it does two unrelated things.  We do want to build this list eventually, for later use by `type_pattern_list`.

In the absence of upstream, I think I would change the name of `vb_pat_constraint` to `find_pat_constraints_and_modes` (or something) and have it call two helpers that do the two things.  But the cost of the extra list is small, and it's nice to stay in sync with upstream, so I'll just do what you suggest (leave `vb_pat_constraint` alone and build an extra list).

typing/typecore.ml
lpw25: That's right. I think that when they differ this upstream code is going to produce confusing error locations.

PR #221, opened 2023-10-25
Ocaml 5 on runtime4: stdlib (except 46aea8c924)

mshinwell:
This should be everything from the stdlib backports, except the stdlib part of `46aea8c924`, which conflicts badly and I will look at separately.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -656,6 +656,7 @@ AS_CASE([$host],
 
 dnl otherlibraries="dynlink runtime_events"
 otherlibraries="dynlink"
+lib_dynlink=true
 AS_IF([test x"$enable_unix_lib" != "xno"],
   [enable_unix_lib=yes
   AC_CONFIG_FILES([otherlibs/unix/META])
--- a/configure.ac
+++ b/configure.ac
@@ -234,6 +234,7 @@ AC_CONFIG_FILES([manual/src/version.tex])
 AC_CONFIG_FILES([manual/src/html_processing/src/common.ml])
 AC_CONFIG_FILES([ocamltest/ocamltest_config.ml])
 AC_CONFIG_FILES([utils/config.generated.ml])
+AC_CONFIG_FILES([tools/eventlog_metadata])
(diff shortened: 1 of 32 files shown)


stdlib/bigarray.mli
```diff
@@ -120,6 +120,7 @@ type ('a, 'b) kind =
   | Complex32 : (Complex.t, complex32_elt) kind
   | Complex64 : (Complex.t, complex64_elt) kind
   | Char : (char, int8_unsigned_elt) kind (**)
+  | Float16 : (float, float16_elt) kind
```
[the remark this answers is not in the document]
xclerc: (If we keep the line, the `(**)` from the
previous line should be moved to that
one.)

configure.ac
```diff
@@ -653,9 +654,8 @@ AS_CASE([$host],
     [exeext=".exe"],
   [exeext=''])
 
-otherlibraries="dynlink runtime_events"
-lib_dynlink=true
```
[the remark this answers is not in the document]
mshinwell: It's unclear this variable is used in the tree, but I've resurrected this line.

configure.ac
```diff
@@ -653,9 +654,9 @@ AS_CASE([$host],
     [exeext=".exe"],
   [exeext=''])
 
-otherlibraries="dynlink runtime_events"
+dnl otherlibraries="dynlink runtime_events"
+otherlibraries="dynlink"
 lib_dynlink=true
-lib_runtime_events=true
```
xclerc, line 658: Sorry, I should have posted that earlier:
I wonder whether it would be safer to
simply `dnl` this line?

xclerc (approved):

configure.ac
mshinwell: Sure, I've left CRs at this point too now.

xclerc: Indeed, I should have suggested
that, without the CRs, the `dnl`
provides little safety.

PR #223, opened 2023-10-25
OCaml 5 on 4 runtime: Backport stdlib pieces of 46aea8c924

ncik-roberts:
Backport the pieces of https://github.com/dra27/ocaml/commit/46aea8c92445a0d7e77db8223b53b83d48fa94c3 that touch files under `stdlib/`. This was done almost entirely by copying-and-pasting from the diff for that commit, as ~everything conflicted.

Backported code is delimited by `BACKPORT BEGIN`/`BACKPORT END` blocks, as in the patch, but I also add a "CR ocaml 5 runtime" needle as we've used elsewhere:

```
(* CR ocaml 5 runtime:
  BACKPORT BEGIN
  <tip-5 code>
*)
<backported code>
(* BACKPORT END *)
```

When code is just commented out and nothing is added in its lieu, that looks like:
```
(* CR ocaml 5 runtime:
  BACKPORT
  <tip-5 code>
*)
```

### DIFF
--- a/stdlib/hashtbl.ml
+++ b/stdlib/hashtbl.ml
@@ -87,7 +87,7 @@ let create ?(random = Atomic.get randomized) initial_size =
     BACKPORT BEGIN
     if random then Random.State.bits (Domain.DLS.get prng_key) else 0
 *)
-    if random then Random.State.bits (Lazy.force prng)
+    if random then Random.State.bits (Lazy.force prng) else 0
 (* BACKPORT END *)
   in
   { initial_size = s; size = 0; seed = seed; data = Array.make s Empty }
--- a/stdlib/hashtbl.ml
+++ b/stdlib/hashtbl.ml
@@ -62,7 +62,12 @@ let randomized = Atomic.make randomized_default
 let randomize () = Atomic.set randomized true
 let is_randomized () = Atomic.get randomized
 
+(* CR ocaml 5 runtime
+   BACKPORT BEGIN
 let prng_key = Domain.DLS.new_key Random.State.make_self_init
+*)
+let prng = lazy (Random.State.make_self_init())
+(* BACKPORT END *)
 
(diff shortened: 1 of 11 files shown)


stdlib/hashtbl.ml
```diff
@@ -62,7 +62,12 @@ let randomized = Atomic.make randomized_default
 let randomize () = Atomic.set randomized true
 let is_randomized () = Atomic.get randomized
 
+(* CR ocaml 5 runtime:
```
xclerc, line 65: We could simplify things a bit by having
```
module Domain = struct
  module DLS = struct
    let new_key ... = ...
    let get ... = ...
  end
end
```
The only `CR ocaml 5 runtime` would
then be to only delete the `Domain`
module.

That being said, I am not positive it is
worth the effort - feel free to resolve
this comment.

stdlib/hashtbl.ml
```diff
@@ -78,7 +83,12 @@ let rec power_2_above x n =
 let create ?(random = Atomic.get randomized) initial_size =
   let s = power_2_above 16 initial_size in
   let seed =
+(* CR ocaml 5 runtime:
+    BACKPORT BEGIN
     if random then Random.State.bits (Domain.DLS.get prng_key) else 0
+*)
+    if random then Random.State.bits (Lazy.force prng)
```
xclerc, line 90: (the `else` arm seems to be missing.)

ncik-roberts: Fixed, thanks.

stdlib/hashtbl.ml
ncik-roberts: I lean toward not doing this because there are only two use sites and it's easy enough to see what's going on.

xclerc (approved):

PR #224, opened 2023-10-25
OCaml 5 on runtime4: asmcomp/ changes + a few others

mshinwell:
(called "backend and lambda" in the internal list, but that is a bit misleading).

This does not include replacing the implementation of "lazy" in matching.ml with the OCaml 4 version.  A separate PR is coming for that.

The changeset that mentions a bootstrap no longer includes one.


xclerc (approved):
I think it is OK to merge; there will be
conflicts with changes in the backend
repository, but we will handle them
when we get there.

PR #225, opened 2023-10-25
OCaml 5 on runtime4: use 4.x lazy implementation

mshinwell:
This just wholesale replaces the implementation rather than using the approach from the backports changesets of commenting out specific sections, which was complicated to look at.  I copy/pasted the 4.x implementation from flambda-backend.

### DIFF
--- a/lambda/matching.ml
+++ b/lambda/matching.ml
@@ -1924,6 +1924,9 @@ let get_mod_field modname field =
          | path, _ -> transl_value_path Loc_unknown env path
        ))
 
+(* BACKPORT BEGIN
+   This is the OCaml 5 lazy implementation merged into ocaml-jst.
+
 let code_force_lazy_block = get_mod_field "CamlinternalLazy" "force_lazy_block"
 
 let code_force_lazy = get_mod_field "CamlinternalLazy" "force_gen"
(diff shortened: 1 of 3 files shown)


ncik-roberts:
Reviewing

lambda/matching.ml
```diff
@@ -2056,6 +2059,141 @@ let inline_lazy_force arg pos loc =
...
+       caml_obj_tag and the switch on arbitrary tags *)
+    (* Lswitch generates compact and efficient native code *)
+    inline_lazy_force_switch arg pos loc
+  else
+    (* generating bytecode: Lswitch would generate too many rather big
+         tables (~ 250 elts); conditionals are better *)
+    inline_lazy_force_cond arg pos loc
+
+
+(* End of lazy implementations. *)
```
ncik-roberts, line 2197: There's a reference to `Obj.forcing_tag` in the implementation of `stdlib/lazy.ml:from_val`, should we delete that too?

ncik-roberts (approved):
I diffed the added code with `tip-jst` and it looks right.

I suggest removing this line from the build script:

```
grep -v "lambda/matching.ml" | # need Obj.forcing_tag in stdlib
```

(I couldn't push to your branch.) I tested this locally and it works.

mshinwell:
> I diffed the added code with `tip-jst` and it looks right.
> 
> I suggest removing this line from the build script:
> 
> ```
> grep -v "lambda/matching.ml" | # need Obj.forcing_tag in stdlib
> ```
> 
> (I couldn't push to your branch.) I tested this locally and it works.

Whoops, fixed

lambda/matching.ml
mshinwell: Ah yes, I've fixed that.  I checked lazy.ml and lazy.mli against the flambda-backend versions and they look ok.

PR #227, opened 2023-10-25
Build everything for OCaml 5 minus

mshinwell:
Contains merges of all outstanding PRs.

PR #233, opened 2023-10-26
Disable tests relying on `Domain`

xclerc:
On top of #227.


ncik-roberts (approved):
Looks good. And I changed the target to `ocaml5-minus-fix-testsuite` so I think OK to merge.

PR #234, opened 2023-10-26
Fixes for Stdlib.Random

mshinwell:
The 5.x `Random` module uses a new PRNG.  I've copied this into `runtime4/misc.c`.  I think this is reasonable even though we are using the 4.x runtime for 5-minus.

### DIFF
--- a/stdlib/random.ml
+++ b/stdlib/random.ml
@@ -113,6 +113,8 @@ module State = struct
           (String.get_int64_le d2 0)
           (String.get_int64_le d2 8)
 
+  let full_init = reinit
+
   let make seed =
     let s = create() in reinit s seed; s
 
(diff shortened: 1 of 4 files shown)


stdlib/random.ml
```diff
@@ -113,6 +113,8 @@ module State = struct
           (String.get_int64_le d2 0)
           (String.get_int64_le d2 8)
 
+  let full_init = reinit
```
ncik-roberts, line 116: It doesn't look like `full_init` has the right type in the mli. (`reinit` takes two arguments, but the mli says `full_init` should have one argument: `val full_init : int array -> unit`.) Is this PR missing something?

ncik-roberts (approved):
Good to merge after resolving the minor comment.

stdlib/random.ml
ncik-roberts: Never mind, I see this is `State.full_init`. This makes sense now.

PR #238, opened 2023-10-26
Remove otherlibs/bigarray/dune

mshinwell:
Stale file.


mshinwell:
One thing we should check is that the `Bigarray` files are still correctly installed (built from `stdlib/` I guess).

xclerc (approved):

mshinwell:
I've made a separate note internally to check build artifacts once flambda-backend builds.

PR #244, opened 2023-10-26
Fix deprecation warnings in st_stubs.c

mshinwell:
Fix from @xclerc

### DIFF
--- a/otherlibs/systhreads/st_stubs.c
+++ b/otherlibs/systhreads/st_stubs.c
@@ -19,6 +19,10 @@
 
 #define CAML_INTERNALS
 
+#define CAML_NAME_SPACE
+#include "caml/compatibility.h"
+#undef CAML_NAME_SPACE
+
 #include "caml/alloc.h"
 #include "caml/backtrace.h"
 #include "caml/callback.h"


ncik-roberts (approved):
Approving to unblock, but I left a comment.

otherlibs/systhreads/st_stubs.c
```diff
@@ -19,6 +19,10 @@
 
 #define CAML_INTERNALS
 
+#define CAML_NAME_SPACE
```
ncik-roberts, line 22: Any reason to prefer this approach over https://github.com/ocaml-flambda/ocaml-jst/commit/c39ae2e6b581e6a6e5649ea329e32a56ad8e2079? (Do we need the linked commit anywhere else?)

mshinwell: Oh, this seems neater?  c39ae2e was a hack that might affect other things.

xclerc: I think it is slightly better because it silences
the deprecation error only in that file, rather
than for all the runtime files.

PR #245, opened 2023-10-26
OCaml 5 testsuite: easy promotions

ncik-roberts:
Promotions involving any of the following:
  * new newline added upstream after printing warnings in the toplevel. I confirmed this is present in ocaml trunk.
  * printing `field_imm` for immutable field accesses
  * changes in the numeric stamps `/[0-9]+` on identifiers in printed lambda. I ran a normalization script locally to confirm that the binding structure is the same after this PR.
  * some extra things we print when printing lambda (`nlocal`, `=[int]` on let-bindings of ints)

Slightly trickier cases:
  * In some places, a `field_mut` in a test added in `tip-5` is changed to `field_imm`. This is because we treat module field accesses as immutable internally and is expected. (Decided with @mshinwell and Vincent Laviron.)
  * A hash type is no longer printed in `testsuite/tests/shapes/simple.ml`. That's expected: see https://github.com/ocaml/ocaml/commit/1632e9b2cfaee20d9921b9e2828abb7931a376ad
  * In `testsuite/tests/tool-toplevel/pr6468.compilers.reference` the difference in stack trace is expected: this was refactored to remove some calls to `Fun.protect`

I have one trivial fix to `printlambda.ml` in this PR to avoid printing `field_mut_mut`.


ccasin (approved):
LGTM

I looked at every diff and confirmed it corresponds to one of the explanations given in the PR description or is otherwise obvious.  I did not independently assess all the given explanations, but they sounds reasonable.

PR #228, opened 2023-10-26
Promote failed tests blindly

ncik-roberts:
Blindly run `promote-failed` based on #227 (07536fcef2ac10a17e957b58a4b88ba64dfbdb95 at time of writing)

PR #258, opened 2023-10-27
Adjust location to match upstream for consistent errors

ccasin:
This fixes [typing-gadts/test.ml](https://github.com/ocaml-flambda/ocaml-jst/pull/228/files#diff-9806dee4a84e77cbd79cba5141ffb88fb80eb8be71f8faa9a38a7b241d33ca0e)

We're parsing type constraints on let bindings a little differently than upstream when they contain locally abstract types (e.g., `let x : type a . a = ...`).  Upstream 5.1 gives the `core_type` node for the annotation a different (probably better) location than it used to, and this simply matches that.

Two other tests are promoted because they contain similar locations:
```
parsetree/locations_test.compilers.reference
typing-layouts/annots_beta.ml
```
In the former case the test also exists upstream, and after this PR we now match the upstream behavior.

Reviewer: @ncik-roberts

### DIFF
--- a/parsing/parser.mly
+++ b/parsing/parser.mly
@@ -636,7 +636,10 @@ let mk_newtypes ~loc newtypes exp =
   in
   List.fold_right mk_one newtypes exp
 
-let wrap_type_annotation ~loc newtypes core_type body =
+(* The [typloc] argument is used to adjust a location for something we're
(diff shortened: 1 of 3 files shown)


ncik-roberts (approved):
I briefly considered whether it's better to have the error location include all of `type (...). ...`, but staying consistent with upstream seems much better.

PR #265, opened 2023-10-30
Pull latest flambda-backend patches

mshinwell:
Prior to the clean subtree import of ocaml-jst into flambda-backend, we need to merge the changes made to the subtree in flambda-backend since we started on the 5-minus work.  These are generated using the `jane/pull-backend-patches` script.  This time there were conflicts and also problems because changes had been made in `runtime/` in flambda-backend that needed to be ported to `runtime4`.  This PR fixes these.


TheNumbat:
I've reviewed the SIMD-related merge (lambda/bytecomp/runtime4 changes) and it looks correct to me.

ncik-roberts:
The commits I pushed are just promoting test failures. This is fine:
  * `ppx-empty-cases`: the tests changed because the lambda we're printing now looks different
  * the location of `Stdlib.failwith` changed in flambda-backend, which affected some backtraces
  * more whitespace was added in some typing error messages

ncik-roberts:
I reviewed using @mshinwell 's suggestion: run the backport script freshly, and look at the diff between the result of that and this PR. After doing that, I'm convinced that Mark's commits only lightly edit the result of the backport script.

ncik-roberts (approved):
I reviewed using @mshinwell 's suggestion: run the backport script freshly, and look at the diff between the result of that and this PR. After doing that, I'm convinced that Mark's commits only lightly edit the result of the backport script, and his changes look reasonable.

PR #266, opened 2023-10-30
OCaml 5: Fix local typing test

ncik-roberts:
Ferry mode-related attributes on value bindings, and have the typechecker copy them to the `Pexp_constraint` and `Ppat_constraint` nodes that it forges.

This is needed because OCaml 5.1 starts parsing `let local_ pat : ty = e` as a value binding with `Pvc_constraint ty`. OCaml 4, in contrast, parses this as a let-binding with `Ppat_constraint (pat, ty)` as the pattern and `Pexp_constraint (e, ty)` as the let-bound expression. We used these two `P*_constraint` nodes internally to hang mode-related attributes off of; now, these nodes don't exist.

Thankfully, there's a dedicated place in the type-checker that creates fake `Pexp_constraint` and `Ppat_constraint` nodes; I can just be careful to copy the mode attributes to the right place there.

This also has a bugfix for a bug introduced in the merge: in `sp = Ppat_constraint (sp', ty)`, we should take the mode annotations from the **parent** node `sp`, not the child node `sp'`. (Check against `tip-jst` if you want to confirm.)

Review suggestion: @ccasin


parsing/builtin_attributes.mli
```diff
@@ -148,9 +155,13 @@ val has_attribute : string list -> Parsetree.attributes -> bool
     "unrolled" only in the case where flambda or flambda2 is configured).  We
     handle this by taking a bool, rather than simply passing fewer nms in those
     cases, to support misplaced attribute warnings - the attribute should not
-    count as misplaced if the compiler could use it in some configuration. *)
+    count as misplaced if the compiler could use it in some configuration.
+
+    [Attributes_filter.t] is an abstract type so that [Builtin_attributes]
+    can expose other filters without calling code inspecting attribute names.
```
ccasin, line 161:
```suggestion
```

(As discussed, I think this comment is more confusing is helpful - this is an abstract type for the usual reasons.)

ccasin (approved):
This looks fine, with one minor suggestion about a comment.

PR #268, opened 2023-10-30
OCaml 5: Backport flambda-backend#1960

ncik-roberts:
Backports ocaml-flambda/flambda-backend#1960.

This should really be based on #265 but I can't do that because it's in a fork. If you're reviewing, you just want to review the last commit.

Review suggestion: @alanechang


alanechang (approved):
The changes look good. Agree the match on the exception tag is not ideal but it does seem like the best thing to do here.

PR #270, opened 2024-07-11
Fix .gitignore, runtime/caml/exec.h* and ocamltest_unix_real.ml

ncik-roberts:
Fix configure script. This is a prerequisite for getting CI to work on my future PRs. The first four commits are copied from https://github.com/mshinwell/ocaml-jst/tree/merge-5.2 -- maybe another PR should be opened with just those commits @mshinwell ?

Changes in my commit:
* Mirrors a change from https://github.com/ocaml/ocaml/pull/12652 onto `runtime4`. I just copied `runtime/caml/exec.h.in` to `runtime4/caml/exec.h.in`; this seems fine because `runtime4/caml/exec.h` and `runtime/caml/exec.h` are identical in flambda-backend.
* Resolve conflicts in `.gitignore` so I can git-ignore the now-auto-generated `runtime4/caml/exec.h` file
* Adopt the upstream naming convention for `ocamltest/ocamltest_unix_real.ml`. For some reason we named them with a `.` internally; we can revisit this later if we run into difficulty, but I think it's OK.


mshinwell:
Yeah I'll open another PR with the first commits tomorrow.
(btw @ncik-roberts `mshinwell/merge-5.2-with-mshinwell-changes` has some more conflict resolution; I will split this into PRs, but check here if you're about to resolve in something other than `parsing/` or `typing/`.)

mshinwell (approved):

PR #271, opened 2024-07-11
Fix parsing mli conflicts, except `parser.mly`

ncik-roberts:
(Only the last 2 commits are to be read. The others are explained in / included in #270 .)

This PR fixes conflicts in parsing mlis and `parser.mly`, and builds them with CI.

Per-file description:
* `parsing/builtin_attributes.mli`: resolve with upstream version, except taking care to keep zero-alloc things. Reasoning: Chris made changes in response to review when posting upstream, and most of those changes apply to flambda-backend too.
* Misc: delete some changes that are duplicated in the merge.
* Misc: disable `flambda2_floats` library for now, as it's not present in ocaml-jst and is breaking the build.

### DIFF
--- a/parsing/parser.mly
+++ b/parsing/parser.mly
@@ -269,34 +269,19 @@ let rec mktailpat nilloc = let open Location in function
 let mkstrexp e attrs =
   { pstr_desc = Pstr_eval (e, attrs); pstr_loc = e.pexp_loc }
 
-<<<<<<< HEAD
-let mkexp_type_constraint ?(ghost=false) ~loc e t =
-  let desc =
-    match t with
-  | N_ary.Pconstraint t -> Pexp_constraint(e, t)
-  | N_ary.Pcoerce(t1, t2)  -> Pexp_coerce(e, t1, t2)
-  in
-  if ghost then ghexp ~loc desc
-  else mkexp ~loc desc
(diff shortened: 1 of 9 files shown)


parsing/parser.mly
```diff
@@ -269,30 +269,15 @@ let rec mktailpat nilloc = let open Location in function
...
-  else mkexp ~loc desc
-||||||| 121bedcfd2
-let mkexp_constraint ~loc e (t1, t2) =
-  match t1, t2 with
-  | Some t, None -> mkexp ~loc (Pexp_constraint(e, t))
-  | _, Some t -> mkexp ~loc (Pexp_coerce(e, t1, t))
-  | None, None -> assert false
-=======
-let mkexp_desc_constraint e t =
+let mkexp_desc_type_constraint e t =
```
ccasin: Why not give this the same name it has upstream (`mkexp_desc_constraint`)?

ccasin (commented):
Everything other than parser.mly looks fine, and I'm pausing review of that to wait for coming nary functions changes.

parsing/parser.mly
ncik-roberts: (Mooted, I removed changes to parser.mly)

PR #273, opened 2024-07-12
Fix conflicts in utils/{clflags,config,misc}.mli

mshinwell:
As subject.


utils/misc.mli
```diff
@@ -165,15 +157,6 @@ module Stdlib : sig
         elements, except for possibly the last list, which may have fewer.
         [chunks_of] raises if [n <= 0]. *)
 
-||||||| 121bedcfd2
-=======
-    val chunks_of : int -> 'a t -> 'a t t
```
xclerc, line 170: Why are we dropping this part?

xclerc (approved):

utils/misc.mli
mshinwell: It's a duplicate (expand the diff just above).

xclerc (approved):

PR #274, opened 2024-07-12
Resolve conflicts in root dune file

mshinwell:
Maybe some more needs doing here, but this should be a good start.

For review the best thing to do is to diff `dune` against `2167f860f7a7396d4a8a119726a1a485960a305d`, which should produce a trivial diff.

### DIFF
--- a/dune
+++ b/dune
@@ -12,11 +12,12 @@
 ;*                                                                        *
 ;**************************************************************************
 
-
 ; set warning as error in release profile
-(env
- (release (flags (:standard -principal -w +a-4-9-40-41-42-44-45-48-66-67-70)))
 
+(env
+ (release
+  (flags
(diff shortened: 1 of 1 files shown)


dune
```diff
@@ -27,380 +28,327 @@
...
-   ;; bytecomp/
-   debug_event meta opcodes bytesections dll symtable
-
-   ;; some of COMP
-   pparse main_args compenv compmisc makedepend compile_common
-   ; manual update: mli only files
-   annot value_rec_types asttypes cmo_format outcometree parsetree debug_event
- ))
+  ;; UTILS
+  config
```
xclerc, line 112: (I think we could make our lives better by
sorting the items, but that's independent
from what we are doing here.)

xclerc (approved):

dune
mshinwell: Yeah indeed

PR #279, opened 2024-07-12
Resolve conflicts in bytecomp/*.mli and remove Symtable.Compunit

mshinwell:
A number of changes here are related to https://github.com/ocaml-flambda/flambda-backend/wiki/Strategy-for-5.2.0-Unit_info-and-compunit .  I've discussed with @lukemaurer .

We have the `Debug_event` interface split out, so I've updated that with changes from `Instruct`.

### DIFF
--- a/bytecomp/bytepackager.mli
+++ b/bytecomp/bytepackager.mli
@@ -23,7 +23,7 @@ type error =
     Forward_reference of string * Compilation_unit.t
   | Multiple_definition of string * Compilation_unit.t
   | Not_an_object_file of string
-  | Illegal_renaming of Compilation_unit.Name.t * string * Compilation_unit.t
+  | Illegal_renaming of Compilation_unit.t * string * Compilation_unit.t
   | File_not_found of string
 
 exception Error of error
--- a/bytecomp/bytepackager.mli
+++ b/bytecomp/bytepackager.mli
@@ -20,16 +20,10 @@ val package_files:
   ppf_dump:Format.formatter -> Env.t -> string list -> string -> unit
 
 type error =
-    Forward_reference of string * Cmo_format.compunit
-  | Multiple_definition of string * Cmo_format.compunit
+    Forward_reference of string * Compilation_unit.t
+  | Multiple_definition of string * Compilation_unit.t
   | Not_an_object_file of string
-<<<<<<< HEAD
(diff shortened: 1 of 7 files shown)


bytecomp/bytepackager.mli
```diff
@@ -20,16 +20,10 @@ val package_files:
...
+  | Multiple_definition of string * Compilation_unit.t
   | Not_an_object_file of string
-<<<<<<< HEAD
-  | Illegal_renaming of Compilation_unit.Name.t * string * string
-||||||| 121bedcfd2
-  | Illegal_renaming of string * string * string
-=======
-  | Illegal_renaming of Cmo_format.compunit * string * Cmo_format.compunit
->>>>>>> 5.2.0
+  | Illegal_renaming of Compilation_unit.Name.t * string * Compilation_unit.t
```
lukemaurer: I suspect what we'll land on here is `Compilation_unit.t * string * Compilation_unit.t` but it's hard to say without seeing the changes we need to make to `bytepackager.ml`.

bytecomp/meta.mli
```diff
@@ -20,14 +20,8 @@ external realloc_global_data : int -> unit = "caml_realloc_global"
...
 external reify_bytecode :
-<<<<<<< HEAD
-  bytes array -> Debug_event.debug_event list array -> string option ->
-||||||| 121bedcfd2
-  bytes array -> Instruct.debug_event list array -> string option ->
-=======
   (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t ->
-  Instruct.debug_event list array -> string option ->
->>>>>>> 5.2.0
+  Debug_event.debug_event list array -> string option ->
```
lukemaurer, line 24: These are the same type, no? Could we use `Instruct.debug_event` here just to keep the diff down?

lukemaurer (approved):
Just a couple of small thoughts, looks good otherwise

bytecomp/meta.mli
mshinwell: I think it's deliberately `Debug_event` to avoid a circular dependency, or similar.  I can't quite remember the details but would rather leave this as-is.

bytecomp/bytepackager.mli
mshinwell: OK I've changed this.

PR #283, opened 2024-07-12
Resolve conflicts in typing/*.mli

mshinwell:
For @ncik-roberts 

I'm reasonably confident about the `Unit_info` changes here.  (See note on https://github.com/ocaml-flambda/flambda-backend/wiki/Strategy-for-5.2.0-Unit_info-and-compunit regarding functions with juxtaposed `Unit_info` and `Compilation_unit` parameters.)

I need to do a bit more work on `typedtree.mli`.

### DIFF
--- a/typing/btype.mli
+++ b/typing/btype.mli
@@ -82,13 +82,7 @@ val is_Tconstr: type_expr -> bool
 val is_Tpoly: type_expr -> bool
 
 val dummy_method: label
-<<<<<<< HEAD
 val type_kind_is_abstract: type_declaration -> bool
-||||||| 121bedcfd2
-=======
-val type_kind_is_abstract: type_declaration -> bool
-val type_origin : type_declaration -> type_origin
->>>>>>> 5.2.0
 
 (**** polymorphic variants ****)
 
--- a/typing/ctype.mli
+++ b/typing/ctype.mli
@@ -176,18 +176,9 @@ val generic_instance: type_expr -> type_expr
 val instance_list: type_expr list -> type_expr list
         (* Take an instance of a list of type schemes *)
 val new_local_type:
-<<<<<<< HEAD
         ?loc:Location.t -> ?manifest_and_scope:(type_expr * int) ->
         Jkind.t -> jkind_annot:Jkind.annotation option -> type_declaration
 val existential_name: constructor_description -> type_expr -> string
-||||||| 121bedcfd2
-        ?loc:Location.t ->
-        ?manifest_and_scope:(type_expr * int) -> unit -> type_declaration
-val existential_name: constructor_description -> type_expr -> string
-=======
-        ?loc:Location.t ->
-        ?manifest_and_scope:(type_expr * int) ->
-        type_origin -> type_declaration
 
 module Pattern_env : sig
   type t = private
@@ -201,7 +192,6 @@ module Pattern_env : sig
   val copy: ?equations_scope:int -> t -> t
   val set_env: t -> Env.t -> unit
 end
->>>>>>> 5.2.0
 
 type existential_treatment =
   | Keep_existentials_flexible
--- a/typing/outcometree.mli
+++ b/typing/outcometree.mli
@@ -58,17 +58,6 @@ type out_value =
   | Oval_variant of string * out_value option
   | Oval_lazy of out_value
 
-<<<<<<< HEAD
-||||||| 121bedcfd2
-type out_type_param = string * (Asttypes.variance * Asttypes.injectivity)
-=======
-type out_type_param = {
-    ot_non_gen: bool;
-    ot_name: string;
-    ot_variance: Asttypes.variance * Asttypes.injectivity
-}
->>>>>>> 5.2.0
-
 type out_modality_legacy = Ogf_global
 
 type out_modality_new = string
@@ -129,6 +118,7 @@ and out_jkind =
 
 and out_type_param =
   { oparam_name : string;
+    oparam_non_gen : bool;
     oparam_variance : Asttypes.variance;
     oparam_injectivity : Asttypes.injectivity;
     oparam_jkind : out_jkind option }
@@ -140,15 +130,9 @@ and out_type =
   | Otyp_abstract
   | Otyp_open
   | Otyp_alias of {non_gen:bool; aliased:out_type; alias:string}
-<<<<<<< HEAD
   | Otyp_arrow of arg_label * out_arg_mode * out_type * out_ret_mode * out_type
   (* INVARIANT: the [out_ret_mode] is [Orm_not_arrow] unless the RHS [out_type]
     is [Otyp_arrow] *)
-||||||| 121bedcfd2
-  | Otyp_arrow of string * out_type * out_type
-=======
-  | Otyp_arrow of Asttypes.arg_label * out_type * out_type
->>>>>>> 5.2.0
   | Otyp_class of out_ident * out_type list
   | Otyp_constr of out_ident * out_type list
   | Otyp_manifest of out_type * out_type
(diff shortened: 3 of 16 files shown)


mshinwell:
Actually maybe `typedtree.mli` is ok.  The diff from flambda-backend is:
```diff
 |and core_type_desc =
 |  | Ttyp_var of string option * Jkind.annotation option
 |  | Ttyp_arrow of arg_label * core_type * core_type
 |  | Ttyp_tuple of (string option * core_type) list
 |  | Ttyp_constr of Path.t * Longident.t loc * core_type list
 |  | Ttyp_object of object_field list * closed_flag
 |  | Ttyp_class of Path.t * Longident.t loc * core_type list
!|  | Ttyp_alias of core_type * string loc option * Jkind.annotation option
 |  | Ttyp_variant of row_field list * closed_flag * label list option
 |  | Ttyp_poly of (string * Jkind.annotation option) list * core_type
 |  | Ttyp_package of package_type
+|  | Ttyp_open of Path.t * Longident.t loc * core_type
 |  | Ttyp_call_pos
 |      (** [Ttyp_call_pos] represents the type of the value of a Position
 |          argument ([lbl:[%call_pos] -> ...]). *)

```
(the change for `Ttyp_alias` is adding the `_ loc`)

typing/types.mli
```diff
@@ -588,14 +582,6 @@ and mixed_product_shape =
...
   }
 
-||||||| 121bedcfd2
-=======
-and type_origin =
-    Definition
-  | Rec_check_regularity       (* See Typedecl.transl_type_decl *)
-  | Existential of string
-
->>>>>>> 5.2.0
```
ncik-roberts, line 598: I suspect we should take the upstream side here. The `abstract_reason` and `type_origin` types are mostly the same, and we should pick the one that reduces drift with upstream.

typing/btype.mli
```diff
@@ -82,13 +82,7 @@ val is_Tconstr: type_expr -> bool
 val is_Tpoly: type_expr -> bool
 
 val dummy_method: label
-<<<<<<< HEAD
 val type_kind_is_abstract: type_declaration -> bool
-||||||| 121bedcfd2
-=======
-val type_kind_is_abstract: type_declaration -> bool
-val type_origin : type_declaration -> type_origin
->>>>>>> 5.2.0
```
ncik-roberts, line 91: We should take upstream.

typing/shape_reduce.mli
```diff
@@ -1,68 +1,3 @@
...
-(** [local_reduce] will not reduce shapes that require loading external
-  compilation units. *)
-val local_reduce : Env.t -> Shape.t -> Shape.t
-
-(** [local_reduce_for_uid] will not reduce shapes that require loading external
-  compilation units. *)
-val local_reduce_for_uid : Env.t -> Shape.t -> result
-||||||| 121bedcfd2
-=======
 (**************************************************************************)
```
ncik-roberts, line 1: Can Luke review this file?

typing/shape.mli
```diff
@@ -49,6 +49,47 @@
     a talk about the reduction strategy
 *)
```
ncik-roberts, line 51: Can Luke review this file?

typing/ctype.mli
```diff
@@ -176,18 +176,9 @@ val generic_instance: type_expr -> type_expr
...
         Jkind.t -> jkind_annot:Jkind.annotation option -> type_declaration
 val existential_name: constructor_description -> type_expr -> string
-||||||| 121bedcfd2
-        ?loc:Location.t ->
-        ?manifest_and_scope:(type_expr * int) -> unit -> type_declaration
-val existential_name: constructor_description -> type_expr -> string
-=======
-        ?loc:Location.t ->
-        ?manifest_and_scope:(type_expr * int) ->
-        type_origin -> type_declaration
```
ncik-roberts, line 190: I'm skeptical about the dropping of `type_origin` and think we should add it to our `new_local_type` function signature for now.

typing/outcometree.mli
```diff
@@ -58,17 +58,6 @@ type out_value =
...
 
-<<<<<<< HEAD
-||||||| 121bedcfd2
-type out_type_param = string * (Asttypes.variance * Asttypes.injectivity)
-=======
-type out_type_param = {
-    ot_non_gen: bool;
-    ot_name: string;
-    ot_variance: Asttypes.variance * Asttypes.injectivity
-}
```
ncik-roberts, line 69: would be better to take the upstream type here to reduce drift, and get rid of our separate `out_type_param` (moving its extra fields to this type)

typing/persistent_env.mli
```diff
@@ -43,29 +43,15 @@ val report_error: Format.formatter -> error -> unit
 module Persistent_signature : sig
```
ncik-roberts, line 43: Can Luke review this file?

ncik-roberts (commented):
There are some changes I'd like to make, but I'd really like to focus on making the parsing change right now. Let me know if you're blocked and I can approve (I will refer back to the comments I made and make a follow-up PR).

typing/shape.mli
```diff
@@ -49,6 +49,47 @@
...
 
+(** A [Uid.t] is associated to every declaration in signatures and
```
lukemaurer, line 52: Looks like this whole comment got duplicated in place?

typing/shape_reduce.mli
```diff
@@ -1,68 +1,3 @@
...
-(*   All rights reserved.  This file is distributed under the terms of    *)
-(*   the GNU Lesser General Public License version 2.1, with the          *)
-(*   special exception on linking described in the file LICENSE.          *)
-(*                                                                        *)
-(**************************************************************************)
-
-(** The result of reducing a shape and looking for its uid *)
-type result =
-  | Resolved of Shape.Uid.t (** Shape reduction succeeded and a uid was found *)
-  | Resolved_alias of Shape.Uid.t list (** Reduction led to an alias chain *)
```
lukemaurer, line 22: Effectively, the only actual diff happening here (I hate add/add conflicts like this) is that we're changing the type of `Resolved_alias` from `Shape.Uid.t list` to `Shape.Uid.t * result`. From what I've seen, it seems like the `list` version is the more convenient API? Are we happy to drop our change? (OTOH odoc's life is always easier if we can drop diffs)

typing/persistent_env.mli
```diff
@@ -106,13 +81,13 @@ type 'a sig_reader =
...
   -> 'a
 
-<<<<<<< HEAD
 (* If [add_binding] is false, reads the signature from the .cmi but does not
    bind the module name in the environment. *)
 (* CR-someday lmaurer: [add_binding] is apparently always false, including in the
    [-instantiate] branch. We should remove this parameter. *)
 val read : 'a t -> 'a sig_reader
-  -> Compilation_unit.Name.t -> filepath -> add_binding:bool -> Subst.Lazy.signature
+  -> Compilation_unit.Name.t -> Unit_info.Artifact.t -> add_binding:bool
```
lukemaurer, line 89: If I'm reading correctly from the conflict above, this should be

```suggestion
  -> Unit_info.Artifact.t -> add_binding:bool
```

(which is to say, the `Unit_info.Artifact.t` is replacing both the `modname` parameter and the `filepath` parameter). Seems reasonable, since the `modname` should in fact be determined by the `filepath` (which is to say, the artifact).

mshinwell: The change was correct as-is I think, as per https://github.com/ocaml-flambda/flambda-backend/wiki/Strategy-for-5.2.0-Unit_info-and-compunit.  Later we can consider putting `Compilation_unit` inside `Unit_info`.

ncik-roberts:
We're closing in favor of a copy to be posted in flambda-backend.

typing/shape_reduce.mli
ncik-roberts: I'm going to make the call to keep the `list` version to avoid coupling this with odoc changes for now.

PR #288, opened 2024-07-15
Make bytecomp/*.ml build

mshinwell:
I've tried to simplify `Bytepackager` as part of changing to use the new types.  @lukemaurer please have an initial look at this.

### DIFF
--- a/bytecomp/bytelink.ml
+++ b/bytecomp/bytelink.ml
@@ -19,17 +19,13 @@ open Misc
 open Config
 open Cmo_format
 
-<<<<<<< HEAD
 module CU = Compilation_unit
 
-||||||| 121bedcfd2
-=======
-module Compunit = Symtable.Compunit
-
->>>>>>> 5.2.0
 module Dep = struct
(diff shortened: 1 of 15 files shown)


lukemaurer:
`Bytepackager` looks sensible to me

bytecomp/bytelink.ml
```diff
@@ -1008,22 +929,11 @@ let report_error ppf = function
...
-        Location.print_filename file2
-        intf
-=======
                  make inconsistent assumptions over interface %a@]"
         (Style.as_inline_code Location.print_filename) file1
         (Style.as_inline_code Location.print_filename) file2
-        Style.inline_code intf
->>>>>>> 5.2.0
+        Style.inline_code
+        (Format.asprintf "%a" CU.Name.print intf)
```
lukemaurer, line 928: It looks from the others like this should be

```suggestion
        (Style.as_inline_code CU.Name.print) intf
```

lukemaurer (approved):
One small thought but otherwise looks good

PR #289, opened 2024-07-15
Make lambda/*.ml build

mshinwell:
As subject.


lambda/translattribute.ml
```diff
@@ -111,45 +46,21 @@ let is_tmc_attribute =
...
+  [ "opaque", Return ]
 
 let is_unboxable_attribute =
-  [ ["unboxable"; "ocaml.unboxable"], true ]
+  [ "unboxable", Return ]
 
 let is_unrolled = function
   | {txt="unrolled"|"ocaml.unrolled"} -> true
   | {txt="inline"|"ocaml.inline"|"inlined"|"ocaml.inlined"} -> false
   | _ -> assert false
```
ncik-roberts: This was removed upstream. Should we try to do the same? (I realize that we, unlike upstream, want to keep around `parse_inlined_attribute` &mdash; which is the only use site of `is_unrolled` &mdash; but maybe we can call `attr_equals_builtin attr "unrolled"` there to match how upstream changed `parse_inline_attribute`?)

lambda/translclass.ml
```diff
@@ -39,28 +39,16 @@ let layout_tables = Lambda.Pvalue Pgenval
...
 let lfunction ?(kind=Curried {nlocal=0}) ?(region=true) ?(ret_mode=alloc_heap) return_layout params body =
   if params = [] then body else
-<<<<<<< HEAD
   match kind, body with
   | Curried {nlocal=0},
     Lfunction {kind = Curried _ as kind; params = params';
                body = body'; attr; loc; mode = Alloc_heap; ret_mode; region}
+    (* XXX mshinwell: 5.2.0 also has:
+         when attr.may_fuse_arity && ...
+    *)
```
ncik-roberts: We should do the same for consistency. (Stephen's RFC doesn't specify syntactic arity for class methods, but the change we made upstream is more consistent with the spirit of the RFC, and we might as well do what we did upstream.)

ncik-roberts (commented):
(still looking, but posting comments so you can see them)

ncik-roberts (approved):

PR #291, opened 2024-07-15
Make Compilenv build

mshinwell:
As subject.  Any further middle-end changes will be fixed in flambda-backend.


mshinwell:
I forgot there is a separate copy of this module in flambda-backend, so will open a separate PR.

PR #292, opened 2024-07-15
Make ocamltest build

mshinwell:
The only notable thing here is that upstream ocamltest now seems to have an explicit stdlib dependency.  I've removed that in this PR so that it can be built using the boot compiler and its stdlib.  (We don't currently require that in flambda-backend I don't think, but it would be useful for this to be possible in future, so tests could be run on the first stage compiler in the event of a suspected miscompilation of the compiler.)


mshinwell:
Moved to flambda-backend

PR #293, opened 2024-07-15
Make toplevel/*.ml build

mshinwell:
As subject.  The more complicated parts here were last touched by the symbols patches I think, so @lukemaurer may be a good choice to review.


lukemaurer (approved):
LGTM

PR #295, opened 2024-07-15
Make driver/*.ml build

mshinwell:
As subject.  The hidden includes changes in `Compmisc` need looking over, I might not have got those correct.  cc @ccasin

### DIFF
--- a/driver/compenv.ml
+++ b/driver/compenv.ml
@@ -714,7 +714,6 @@ let process_deferred_actions env =
     end
   else if !deferred_actions = [] then
     fatal "No input files";
-  (* XXX mshinwell: do we need this? *)
   let compiling_multiple_impls =
     List.length (List.filter (function
         | ProcessImplementation _ -> true
--- a/driver/compenv.ml
+++ b/driver/compenv.ml
@@ -62,84 +62,6 @@ let first_objfiles = ref []
 let last_objfiles = ref []
 let stop_early = ref false
 
-<<<<<<< HEAD
-(* Check validity of module name *)
-let is_unit_name name =
-  try
-    if name = "" then raise Exit;
-    begin match name.[0] with
-    | 'A'..'Z' -> ()
-    | _ ->
-       raise Exit;
-    end;
-    for i = 1 to String.length name - 1 do
-      match name.[i] with
-      | 'A'..'Z' | 'a'..'z' | '0'..'9' | '_' | '\'' -> ()
-      | _ ->
-         raise Exit;
-    done;
-    true
-  with Exit -> false
-
-let check_unit_name filename name =
-  if not (is_unit_name name) then
-    Location.prerr_warning (Location.in_file filename)
-      (Warnings.Bad_module_name name)
-
-(* Compute name of module from output file name *)
-let module_of_filename inputfile outputprefix =
-  let basename = Filename.basename outputprefix in
-  let name =
-    try
-      (* For hidden files (i.e., those starting with '.'), include the initial
-         '.' in the module name rather than let it be empty. It's still not a
-         /good/ module name, but at least it's not rejected out of hand by
-         [Compilation_unit.Name.of_string]. *)
-      let pos = String.index_from basename 1 '.' in
-      String.sub basename 0 pos
-    with Not_found -> basename
-  in
-  let name = String.capitalize_ascii name in
-  check_unit_name inputfile name;
-  name
-||||||| 121bedcfd2
-(* Check validity of module name *)
-let is_unit_name name =
-  try
-    if name = "" then raise Exit;
-    begin match name.[0] with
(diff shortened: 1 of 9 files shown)


driver/compenv.ml
```diff
@@ -62,84 +62,6 @@ let first_objfiles = ref []
...
-(* Compute name of module from output file name *)
-let module_of_filename inputfile outputprefix =
-  let basename = Filename.basename outputprefix in
-  let name =
-    try
-      (* For hidden files (i.e., those starting with '.'), include the initial
-         '.' in the module name rather than let it be empty. It's still not a
-         /good/ module name, but at least it's not rejected out of hand by
-         [Compilation_unit.Name.of_string]. *)
-      let pos = String.index_from basename 1 '.' in
```
ccasin, line 98: This bit is the source of the conflict.  Is some equivalent code needed in a new place (`Unit_info`, maybe) to be compatible with @lukemaurer's symbols changes?  I haven't traced this through all the way - I imagine Luke will know immediately.

driver/compenv.ml
```diff
@@ -818,8 +714,15 @@ let process_deferred_actions env =
     end
   else if !deferred_actions = [] then
     fatal "No input files";
-  List.iter (process_action env) (List.rev !deferred_actions);
->>>>>>> 5.2.0
+  (* XXX mshinwell: do we need this? *)
```
ccasin: (Just noting I did not assess this question as part of review - let me know if you expected me to.)

driver/compile_common.ml
```diff
@@ -16,17 +16,8 @@
...
-  module_name : Compilation_unit.t;
-  output_prefix : string;
-||||||| 121bedcfd2
-  source_file : string;
-  module_name : string;
-  output_prefix : string;
-=======
   target: Unit_info.t;
->>>>>>> 5.2.0
+  module_name : Compilation_unit.t;
```
ccasin, line 20: This should probably replace `modname` in `Unit_info.t` and be removed from here?

driver/compmisc.ml
```diff
@@ -26,6 +26,9 @@ let auto_include find_in_dir fn =
    then the directories specified with the -I option (in command line order),
    then the standard library directory (unless the -nostdlib option is given),
    then the directories specified with the -H option (in command line order).
+   then the directories specified with the -I option (in command line order),
+   then the standard library directory (unless the -nostdlib option is given),
+   then the directories specified with the -H option (in command line order).
```
ccasin: Looks like an accidental comment change?

driver/compmisc.ml
```diff
@@ -41,31 +44,17 @@ let init_path ?(auto_include=auto_include) ?(dir="") () =
...
-  in
   let visible =
     (if !Clflags.no_cwd then [] else [dir])
     @ List.rev_append visible (Clflags.std_include_dir ())
   in
-<<<<<<< HEAD
   let hidden =
     List.rev_map (Misc.expand_directory Config.standard_library)
       !Clflags.hidden_include_dirs
+    @ List.rev_append visible (Clflags.std_include_dir ())
```
ccasin: I don't quite understand what happened in the merge resolution here - I think you want to keep the deleted lines 44-46 and then take the HEAD version here unchanged.

ccasin (commented):
I looked at `Compmisc`, as requested.  I then started reading the rest (some comments in `Compenv` and `Compiler_common`), but I think @lukemaurer would be best to review the rest of these changes, which are mostly conflict resolutions for PRs of his.  I can dig in if he's unavailable, though.

driver/compile_common.ml
mshinwell: Yes, but in due course.  See https://github.com/ocaml-flambda/flambda-backend/wiki/Strategy-for-5.2.0-Unit_info-and-compunit

driver/compenv.ml
lukemaurer: Ah—yes, we need to change `Unit_info.basename_chop_extensions` so that it calls `String.index_from` instead of `String.index`. I think it all hangs together besides that (thankfully, there's surprisingly little overlap between `Unit_info` and `Compilation_unit`)

driver/compile_common.ml
ccasin: OK.  I worry a little bit that having `modname` sitting over there could re-introduce whatever issues `Compilation_unit.t` was introduced to solve, but I defer to you and Luke here.

mshinwell:
@ccasin @lukemaurer please look again now, thanks

ccasin (approved):
Approving for compmisc, assume Luke will look at the rest.

lukemaurer (approved):

lukemaurer (approved):

mshinwell:
Moved to flambda-backend

